PR #56, opened 2022-07-05
Updated Devouring Plague for spriest

dbyena:
Updated devouring plague spell. This will not be the final update of the spell, but I made a note for what needs to be changed.


dbyena:
I also need to figure out a way to handle the BonusSpellHitRating misery gives when the debuff is active. This will need to be for all spells when misery is active.

sim/priest/devouring_plague.go
```diff
@@ -24,43 +36,45 @@ func (priest *Priest) registerDevouringPlagueSpell() {
...
 				core.TernaryFloat64(priest.Talents.Shadowform, 1.15, 1),
 			ThreatMultiplier: 1 - 0.08*float64(priest.Talents.ShadowAffinity),
 			IsPeriodic:       true,
-			BaseDamage:       core.BaseDamageConfigMagicNoRoll(1216/8, 0.1),
-			OutcomeApplier:   priest.OutcomeFuncTick(),
+			BaseDamage:       core.BaseDamageConfigMagicNoRoll(1376/8, 0.1849),
+
+			// OutcomeApplier: priest.DPcrits()
+			// Assume shadow form is always active for right now
+			//if priest.Talents.Shadowform > 0 { // NEED TO ADD A CHECK TO SEE IF SHADOWFORM TALENT IS CHOSEN, THEN DETERMINE IF DOTS CAN CRIT OR NOT
```
jimmyt857, line 65: The way you can do this is to create a variable outside this definition and then use it here. E.g:

```
outcomeApplier := priest.OutcomeFuncTick()
if priest.Talents.Shadowform {
  outcomeApplier = priest.OutcomeFuncMagicCrit(priest.SpellCritMultiplier(1, 1))
}

...

priest.DevouringPlagueDot = ...
  ...
  OutcomeApplier: outcomeApplier,
  ...
}
```

dbyena:
Need to also increase critical strike chance of the dots with mindmelt talent.

sim/priest/devouring_plague.go
```diff
@@ -47,31 +52,24 @@ func (priest *Priest) registerDevouringPlagueSpell() {
 			ActionID: actionID,
 		}),
 
-		NumberOfTicks: 8,
-		TickLength:    time.Second * priest.ApplyCastSpeed(3),
-
+		NumberOfTicks:       8,
+		TickLength:          time.Second * priest.ApplyCastSpeed(3),
```
lologarithm, line 56: i think you can use the `AffectedByCastSpeed: true` and then set the tick length to just `time.Second*3`

PR #62, opened 2022-07-06
Update armor penetration calculations and debuffs

b-r-a-n:
1. Introduces a PseudoStat for ArmorMultiplier
2. Add Armor() and ArmorPenetration() functions to unit that can be used to get the modified armor / armor pen percentages respectively
3. Updates debuffs to modify this multiplier and be tagged as Major/Minor (20%/5%)
4. Updates the damage reduction formula based on the model that has been discussed in the simonize vid
5. Adds tests to verify damage reduction and aura stacking

Still missing:
1. Correct spell ids for debuffs (only updated their respective armor modifiers)
2. Talent and stance based armor pen

### DIFF
--- a/sim/core/armor_test.go
+++ b/sim/core/armor_test.go
@@ -1,0 +1,264 @@
+package core
+
+import (
+	"testing"
+
+	"github.com/wowsims/wotlk/sim/core/stats"
+)
+
+func TestSunderArmorStacks(t *testing.T) {
+	sim := Simulation{}
+	baseArmor := 10643.0
+	target := Unit{
+		Type:         EnemyUnit,
+		Index:        0,
+		Level:        83,
+		auraTracker:  newAuraTracker(),
+		initialStats: stats.Stats{stats.Armor: baseArmor},
+		PseudoStats:  stats.NewPseudoStats(),
+		Metrics:      NewUnitMetrics(),
+	}
+	target.stats = target.initialStats
+	expectedArmor := baseArmor
+	if target.Armor() != expectedArmor {
+		t.Fatalf("Armor value for target should be %f but found %f", 10643.0, target.Armor())
+	}
+	stacks := int32(1)
+	sunderAura := SunderArmorAura(&target, stacks)
+	sunderAura.Activate(&sim)
+	tolerance := 0.001
+	for stacks <= 5 {
+		expectedArmor = baseArmor * (1.0 - float64(stacks)*0.04)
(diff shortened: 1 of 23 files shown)


sim/core/armor_test.go
```diff
@@ -0,0 +1,264 @@
...
+	if target.Armor() != expectedArmor {
+		t.Fatalf("Armor value for target should be %f but found %f", 10643.0, target.Armor())
+	}
+	stacks := int32(1)
+	sunderAura := SunderArmorAura(&target, stacks)
+	sunderAura.Activate(&sim)
+	tolerance := 0.001
+	for stacks <= 5 {
+		expectedArmor = baseArmor * (1.0 - float64(stacks)*0.04)
+		if target.Armor() < expectedArmor-tolerance || target.Armor() > expectedArmor+tolerance {
```
jimmyt857, line 32: This tolerance logic is used a lot. Could it be a new function in sim/core/utils.go, and maybe define tolerance just once as a const at the top of this file?

sim/core/debuffs.go
```diff
@@ -453,26 +455,71 @@ func ScheduledSunderArmorAura(target *Unit) *Aura {
...
+		},
+		OnStacksChange: func(aura *Aura, sim *Simulation, oldStacks int32, newStacks int32) {
+			oldMultiplier := (1.0 - float64(oldStacks)*armorReductionPerStack)
+			newMultiplier := (1.0 - float64(newStacks)*armorReductionPerStack)
+			aura.Unit.PseudoStats.ArmorMultiplier *= (newMultiplier / oldMultiplier)
+		},
+	})
+}
+
+func ScheduledAcidSpitAura(target *Unit) *Aura {
```
jimmyt857, line 481: Instead of defining a new function, lets make a ScheduledAura() function which takes additional parameters for the aura and time to apply.

sim/core/armor_test.go
b-r-a-n: Updated the PR to add this function and use it in the armor tests. I can refactor the other places in another PR.

sim/core/debuffs.go
b-r-a-n: Made this change but may need some tweaks. I'll leave the comment unresolved until I get some feedback on it. Specifically, not sure whether it makes sense to use the options structure as a parameter (as I have currently). Primary reason I chose to was that the OnAction handler is different but that difference seems like it could be inferred (in all of the current use cases).

PR #63, opened 2022-07-06
Updated Mind Blast and Shadow Word Death Spells

dbyena:
Updated mind blast and shadow word death priest spells. Still need to update including replenishment on MB hits and glyph of shadow word death if target is 35% health, but these are minor updates that can be made later.


dbyena:
The test errors have something to do with assuming misery is always active, so I couldnt just check if active(), but I had to check if not nil.

dbyena:
Please merge when you can as I will not be available tomorrow. Thanks.

PR #74, opened 2022-07-08
update priest mind flay and a few talents

CyberQuackGames:
Mainly using same updates used for MB for MF here. Also went through and cleaned up a few talent references for spriest. We still need to follow up with dot refresh logic after SWP updates.


CyberQuackGames:
Something odd going on with `AllItems-TheLightningCapacitor-28785` test here. Doing the standard `make test && make update-tests` didn't seem to work if the tests are run again after.

PR #99, opened 2022-07-11
Mostly fix armor calcs

jimmyt857:


sim/core/armor_test.go
```diff
@@ -224,39 +224,39 @@ func TestDamageReductionFromArmor(t *testing.T) {
...
+	expectedDamageReduction = 0.6532
 	if !WithinToleranceFloat64(expectedDamageReduction, attackTable.ArmorDamageReduction, tolerance) {
 		t.Fatalf("Expected major & minor armor modifier to result in %f damage reduction got %f", expectedDamageReduction, attackTable.ArmorDamageReduction)
 	}
 
 	// Cap armor pen
 	attacker.stats[stats.ArmorPenetration] = 1400
 	attackTable.UpdateArmorDamageReduction()
-	expectedDamageReduction = 0.0203
+	expectedDamageReduction = 0.9797
```
b-r-a-n, line 235: The new test results don't seem intuitive. An attacker with 1400 armor pen and all the debuffs should be around 2% damage reduction from armor. The values we are getting from the test now are just 1 - the original value. This makes sense based on the change to the implementation but doesn't follow the damage reduction semantics it should.

sim/core/spell_resistances.go
```diff
@@ -45,9 +45,11 @@ func (spellEffect *SpellEffect) applyResistances(sim *Simulation, spell *Spell,
...
 // ArmorDamageReduction currently assumes a level 80 attacker
 func (at *AttackTable) UpdateArmorDamageReduction() {
-	reducibleArmor := MinFloat((at.Defender.Armor()+ReducibleArmorConstant)/3, at.Defender.Armor())
-	effectiveArmor := MaxFloat(at.Defender.Armor()-reducibleArmor*at.Attacker.ArmorPenetration(), 0)
-	at.ArmorDamageReduction = 1 / ((float64(at.Attacker.Level)*467.5-22167.5)/effectiveArmor + 1)
+	defenderArmor := at.Defender.Armor()
+	reducibleArmor := MinFloat((defenderArmor+ReducibleArmorConstant)/3, defenderArmor)
+	effectiveArmor := defenderArmor - reducibleArmor*at.Attacker.ArmorPenetration()
+	armorConstant := float64(at.Attacker.Level)*467.5 - 22167.5
+	at.ArmorDamageReduction = 1 - effectiveArmor/(effectiveArmor+armorConstant)
```
b-r-a-n, line 52: This is basically 1 - previous implementation value (I think). I think changing this in spell_resistances.go (to be 1- damage reduction) is a better fix.

PR #102, opened 2022-07-11
Initial DK pull request

JooperGH:


sim/core/rune.go
```diff
@@ -0,0 +1,208 @@
...
+			Cooldown{unit.NewTimer(), 10 * time.Second},
+			Cooldown{unit.NewTimer(), 10 * time.Second},
+			Cooldown{unit.NewTimer(), 10 * time.Second},
+		},
+
+		onRuneGain: onDeathRuneGain,
+	}
+}
+
+func (unit *Unit) HasBloodRuneBar() bool {
```
jimmyt857, line 131: It should be fine to have only 1 of these functions (i.e. HasRunicPower) since they're all connected.

sim/deathknight/deathknight.go
```diff
@@ -0,0 +1,286 @@
...
+		ModifiedStat: stats.AttackPower,
+		Modifier: func(strength float64, attackPower float64) float64 {
+			return attackPower + strength*2
+		},
+	})
+
+	return deathKnight
+}
+
+func RegisterDeathKnight() {
```
jimmyt857, line 148: Put this at the top of the file if you don't mind, to be consistent with the other classes.

ui/deathknight/presets.ts
```diff
@@ -0,0 +1,140 @@
...
+	name: 'Frost Dps',
+	data: '23050005-32005350352203012300033101351',
+};
+
+export const DefaultRotation = DeathKnightRotation.create({
+	useScourgeStrike: false,
+});
+
+export const DefaultOptions = DeathKnightOptions.create({
+	dualWhield: true,
```
jimmyt857, line 39: I think this field is spelled wrong; also dual wielding is implicit depending on whether the user selected 2 weapons, so you don't need an option for it.

wotlk-deathknight.code-workspace
```diff
@@ -0,0 +1,8 @@
+{
```
jimmyt857, line 1: I don't think you're supposed to commit this file  :)

Also looks like a `wowsimwotlk` binary file is being committed, that should be removed too.

sim/core/stats/stats.go
```diff
@@ -44,6 +44,11 @@ const (
 	Mana
 	Energy
 	Rage
+	RunicPower
```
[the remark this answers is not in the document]
JooperGH: I'll delete it

sim/deathknight/deathknight.go
JooperGH: Done

PR #106, opened 2022-07-12
Shadow priest wrath rotation updates

dbyena:
Updated Spriest rotations for basic/clipping/ideal. Clipping/basic need minor updates still, but this is a good place to start for wrath analysis. Ideal is pretty much there, but definitely some work to do on snap shotting, but most of the stuff is in.


dbyena:
Accidently merged 2 pull requests

PR #117, opened 2022-07-12
Big rune overhaul, and some talents + skills

JooperGH:


.gitignore
```diff
@@ -8,6 +8,7 @@ node_modules
 # binaries
 dist
 binary_dist
+sim/web
```
lologarithm, line 11: We don't want to ignore the whole directory, just the file. I actually just added this change in my recent PR too. You can probably just remove this one :)

proto/common.proto
```diff
@@ -417,7 +417,7 @@ message RaidBuffs {
 	TristateEffect strength_of_earth_totem = 14; // none, normal, 15% improved
 	
 	// 20% Melee Haste
-	TristateEffect icy_talons = 15;
+	bool icy_talons = 15;
```
lologarithm, line 420: Ahh, I thought icy talons was like WF which can be 16% or 20%. I will fix this in my next PR

lologarithm: ah nvermind, you already fixed it :)

PR #127, opened 2022-07-14
added glyphs and set bonuses to rotation code

dbyena:
added glyphs and set bonuses to rotation code

### DIFF
--- a/sim/priest/shadow/rotation.go
+++ b/sim/priest/shadow/rotation.go
@@ -1,13 +1,14 @@
 package shadow
 
 import (
+	"math"
+	"time"
 	//"sort"
+
 	"github.com/wowsims/wotlk/sim/core"
 	"github.com/wowsims/wotlk/sim/core/proto"
 	"github.com/wowsims/wotlk/sim/core/stats"
 	"github.com/wowsims/wotlk/sim/priest"
-	"math"
-	"time"
 )
 
 // TODO: probably do something different instead of making it global?
@@ -73,11 +74,11 @@ func (spriest *ShadowPriest) tryUseGCD(sim *core.Simulation) {
 	if spriest.HasSetBonus(priest.ItemSetCrimsonAcolyte, 4) {
 		mf_reduc_time = time.Millisecond * 510
 	}
-	tickLength := spriest.ApplyCastSpeed(time.Second) - mf_reduc_time
+	tickLength := spriest.ApplyCastSpeed(time.Second - mf_reduc_time)
 	DotTickSpeed := float64(spriest.ApplyCastSpeed(time.Second * 3))
 	critChance := (spriest.GetStat(stats.SpellCrit) / (core.CritRatingPerCritChance * 100))
 	remain_fight := float64(sim.GetRemainingDuration())
-	bosshealth := float64(sim.GetRemainingDurationPercent())
+	//bosshealth := float64(sim.GetRemainingDurationPercent())
 	castmf2 := 0 // if SW stacks = 3, and we want to get SWP up at 5 stacks exactly, then we want to hard code a MF2
 	bestIdx := -1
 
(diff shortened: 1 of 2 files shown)


sim/priest/shadow/rotation.go
```diff
@@ -1,12 +1,13 @@
...
 
 import (
-	"time"
 	//"sort"
 	"github.com/wowsims/wotlk/sim/core"
 	"github.com/wowsims/wotlk/sim/core/proto"
 	"github.com/wowsims/wotlk/sim/core/stats"
+	"github.com/wowsims/wotlk/sim/priest"
 	"math"
+	"time"
```
jimmyt857, line 10: We keep the standard library imports separate from the github imports, with 1 blank line in between. So do:

```
math
time

github/...
github/...
```

sim/priest/shadow/rotation.go
```diff
@@ -68,10 +69,15 @@ func (spriest *ShadowPriest) tryUseGCD(sim *core.Simulation) {
 	// initialize helpful variables for calculations later
 	vtCastTime := spriest.ApplyCastSpeed(time.Millisecond * 1500)
 	gcd := spriest.SpellGCD()
-	tickLength := spriest.ApplyCastSpeed(time.Second)
+	var mf_reduc_time time.Duration
+	if spriest.HasSetBonus(priest.ItemSetCrimsonAcolyte, 4) {
```
jimmyt857, line 73: It might be cleaner to have a helper function in mind_flay.go which returns the base channel tick time. Then you could use that here and also in the construction of the Mind Flay spell.

sim/priest/shadow/rotation.go
```diff
@@ -68,10 +69,15 @@ func (spriest *ShadowPriest) tryUseGCD(sim *core.Simulation) {
...
+		mf_reduc_time = time.Millisecond * 510
+	}
+	tickLength := spriest.ApplyCastSpeed(time.Second) - mf_reduc_time
 	DotTickSpeed := float64(spriest.ApplyCastSpeed(time.Second * 3))
 	critChance := (spriest.GetStat(stats.SpellCrit) / (core.CritRatingPerCritChance * 100))
 	remain_fight := float64(sim.GetRemainingDuration())
+	bosshealth := float64(sim.GetRemainingDurationPercent())
```
jimmyt857, line 80: We have a built-in mechanism for this, which you need to use in order to work properly with the user inputs in the UI. Call this function to check for 35% execute: https://github.com/wowsims/wotlk/blob/master/sim/core/sim.go#L363

sim/priest/shadow/rotation.go
```diff
@@ -150,6 +156,14 @@ func (spriest *ShadowPriest) tryUseGCD(sim *core.Simulation) {
 			TFmod = 0
 		}
 
+		mfglyphMod := 1.0
+		if spriest.HasGlyph(int32(proto.PriestMajorGlyph_GlyphOfMindFlay)) {
```
jimmyt857, line 160: Instead of casting all of these glyphs, create helper functions for priest, analogous to https://github.com/wowsims/wotlk/blob/master/sim/hunter/hunter.go#L84

sim/priest/shadow/rotation.go
dbyena: I'll work on adding these optimizations later in code clean up, but I am trying to figure out core rotation stuff first and I suck at minor things like this.

sim/priest/shadow/rotation.go
dbyena: The helper function seems pretty close to what is already in this code. Small implementation differences, but seems fine. Maybe im not understanding.

sim/priest/shadow/rotation.go
dbyena: updated

sim/priest/shadow/rotation.go
dbyena: updated

PR #177, opened 2022-07-18
Update RegEx for prismatic gems

EsreverWoW:
Currently missing gems such as Enchanted Tear / Nightmare Tear / etc.

### DIFF
--- a/generate_items/wowhead_api.go
+++ b/generate_items/wowhead_api.go
@@ -524,7 +524,7 @@ var gemSocketColorPatterns = map[proto.GemColor]*regexp.Regexp{
 	proto.GemColor_GemColorOrange:    regexp.MustCompile("Matches a ((Yellow)|(Red)) or ((Yellow)|(Red)) (S|s)ocket\\."),
 	proto.GemColor_GemColorPurple:    regexp.MustCompile("Matches a ((Blue)|(Red)) or ((Blue)|(Red)) (S|s)ocket\\."),
 	proto.GemColor_GemColorGreen:     regexp.MustCompile("Matches a ((Yellow)|(Blue)) or ((Yellow)|(Blue)) (S|s)ocket\\."),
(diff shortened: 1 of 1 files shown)


jimmyt857:
Thanks for the fix! You should run `make items` in your local branch to generate the new gems, and include it in this PR.

generate_items/wowhead_api.go
```diff
@@ -524,7 +524,7 @@ var gemSocketColorPatterns = map[proto.GemColor]*regexp.Regexp{
 	proto.GemColor_GemColorOrange:    regexp.MustCompile("Matches a ((Yellow)|(Red)) or ((Yellow)|(Red)) (S|s)ocket\\."),
 	proto.GemColor_GemColorPurple:    regexp.MustCompile("Matches a ((Blue)|(Red)) or ((Blue)|(Red)) (S|s)ocket\\."),
 	proto.GemColor_GemColorGreen:     regexp.MustCompile("Matches a ((Yellow)|(Blue)) or ((Yellow)|(Blue)) (S|s)ocket\\."),
-	proto.GemColor_GemColorPrismatic: regexp.MustCompile("Matches a Red, Yellow or Blue (S|s)ocket\\."),
+	proto.GemColor_GemColorPrismatic: regexp.MustCompile("Matches a Red, Yellow or Blue (S|s)ocket\\."|"Matches any socket\\."|"Matches any Socket\\.  Maximum of one socketed in your equipment"),
```
lologarithm, line 527: get rid of the " marks in there. Should just be around the string itself

lologarithm: and probably dont need `Matches any Socket\\.  Maximum of one socketed in your equipment` at the end since the first one `Matches any socket\\.` would have already caught it.

lologarithm:
looks like we already integrated this fix. Thanks for the PR Esrever

PR #196, opened 2022-07-19
Add webhook actions

Kruhlmann:
Adds two new GitHub actions:

**on_issue**

Triggers a webhook in the Discord server, whenever a new issue is created.

**on_pull_request**

Triggers a webhook in the Discord server, whenever a new pull request is opened on master.

Requires a new secret, `WEBHOOK_URL` (do not append "/github" to the end of the URL, that's done by the action by default).

### DIFF
--- a/.github/workflows/on_issue.yml
+++ b/.github/workflows/on_issue.yml
@@ -8,5 +8,8 @@ jobs:
     steps:
       - name: Send webhook
         uses: Ilshidur/action-discord@master
+        continue-on-error: true
         env:
           DISCORD_WEBHOOK: ${{ secrets.WEBHOOK_URL }}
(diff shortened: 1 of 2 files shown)


jimmyt857:
To clarify, should this be merged? Or are we still waiting for something?

Kruhlmann:
PRs should not depend on the webhook action to succeed in order to merge. We still need to ignore it the PR checks. I'll have a look at that.

Kruhlmann:
Since the webhook only triggers whenver a new PR is opened we won't be able to see it here. I've added `continue_on_error: true` to the step and added another, which simply exits with `0`. This should stop the action from blocking PRs from foreign branches.

PR #211, opened 2022-07-21
Enhancement Shaman Changes

Wumblesnarf:
Imbues implemented, including glyphs (except windfury, dont know how it interacts with dual wield yet) and elemental t4, fixed stormstrike from infinitely scaling (still needs a better solution but this works for the time being), some cleanup of tbc leftovers like rockbiter and setting primary shock


Watcher7:
LGTM

I just have a question for @jimmyt857 
Is the `package-lock.json` file to be considered frozen unless updated by head developers?
I've made a point of excluding it from my own commit, because I noticed everyone else did.
But I didn't see any hard rules on that-- nor anything like a CODEOWNERS for it.

Wumblesnarf:
I can start excluding the file, I just didnt know there was a convention around it haha, still pretty new to this

PR #241, opened 2022-07-23
Warrior Pre-Alpha

TheGroxEmpire:
**Warning**: I have to remove https://github.com/wowsims/wotlk/blob/master/ui/core/proto_utils/gear.ts#L74 for titan grips as I can't access player spec in here.

Known Issue:
- Rotation is not optimal
- Rend debuff refresh is not being prioritized correctly
- Indestructible Potion is not in the UI option yet, but is applied and working on the default option
- Slam proc for fury is not working yet
- Rend doesn't do extra 35% damage when target health is >75%


sim/core/stats/stats.go
```diff
@@ -311,16 +311,17 @@ type PseudoStats struct {
...
-	BonusOHArmorPenRating float64 // Talents, e.g. Rogue Mace specialization
-	BonusSpellCritRating  float64 // Crit rating bonus to spells
+	BonusRangedHitRating      float64 // Hit rating for ranged only.
+	BonusMeleeCritRating      float64 // Crit rating for melee only (not ranged).
+	BonusRangedCritRating     float64 // Crit rating for ranged only.
+	BonusFireCritRating       float64 // Crit rating for fire spells only.
+	BonusShadowCritRating     float64 // Crit rating for shadow spells only. Warlock stuff. You wouldn't understand.
+	BonusMHCritRating         float64 // Talents, e.g. Rogue Dagger specialization
+	BonusOHCritRating         float64 // Talents, e.g. Rogue Dagger specialization
+	BonusMeleeSpellCritRating float64 // Crit rating for melee special attacks, used for Warrior Recklessness
```
rosenrusinov, line 321: If this stat is used only for warrior then it might not be a good idea to be a PseudoStat. 

Another way to implement this is to add a custom wrapper to the OutcomeApplier and modify the crit rating in there depending if reck is active or not

PR #245, opened 2022-07-23
Propose MW3/4 LB weaving implementation

lantisnt:
Preppared space to add MW3/4 weave param


Wumblesnarf:
im not sure if we'd want MW3 and MW4 weaves at that point in the prio, probably need something that allows a more configurable spell prio as certain things can change that, also a no-weave rotation will probably be needed considering the way ive seen some people talk about weaving

PR #272, opened 2022-07-24
WARLOCK HYPE

Glandaaalf:


sim/core/buffs.go
```diff
@@ -55,9 +55,11 @@ func applyBuffEffects(agent Agent, raidBuffs proto.RaidBuffs, partyBuffs proto.P
...
 	if raidBuffs.MoonkinAura == proto.TristateEffect_TristateEffectImproved || raidBuffs.SwiftRetribution {
-		character.PseudoStats.CastSpeedMultiplier *= 1.03
-		character.PseudoStats.MeleeSpeedMultiplier *= 1.03
-		character.PseudoStats.RangedSpeedMultiplier *= 1.03
+		hastePercentBonus := 3.0
+		character.AddStats(stats.Stats{
+			stats.SpellHaste: hastePercentBonus*HasteRatingPerHastePercent,
+			stats.MeleeHaste: hastePercentBonus*HasteRatingPerHastePercent,
+		})
 	}
```
rosenrusinov, line 63: This breaks DKs ghoul pet scaling as it scales with its owners haste rating. Having the buff give haste rating means the pet will first get the the bonus from the buff on the owner then get it again when the buff is applied to it (permanent pets receive raid buffs)

Why did you need to make this change?

Why do you need this change?

Glandaaalf: I needed to change it because it now fits the buff description and now you can actually correctly see it in the UI.
Sadly, there's no way around that since "CastSpeedMultiplier" is different from Haste.

jimmyt857: Don't change internals like this just to satisfy something in the UI; this is going to change actual DPS values, as the haste stat and speed multipliers do not work the same way.

We can chat about adding something to make this visible in UI but it should be separate.

sim/core/buffs.go
```diff
@@ -197,7 +199,9 @@ func applyBuffEffects(agent Agent, raidBuffs proto.RaidBuffs, partyBuffs proto.P
 		})
 	}
 	if raidBuffs.WrathOfAirTotem {
-		character.PseudoStats.CastSpeedMultiplier *= 1.05
+		character.AddStats(stats.Stats{
+			stats.SpellHaste: 5.0*HasteRatingPerHastePercent,
```
jimmyt857, line 203: This is a DPS change for all classes, I think it's incorrect unless you have some data backing it up.

sim/core/buffs.go
```diff
@@ -226,9 +230,13 @@ func applyBuffEffects(agent Agent, raidBuffs proto.RaidBuffs, partyBuffs proto.P
 	}
 
 	if raidBuffs.IcyTalons {
-		character.PseudoStats.MeleeSpeedMultiplier *= 1.2
+		character.AddStats(stats.Stats{
+			stats.MeleeHaste: 20*HasteRatingPerHastePercent,
```
jimmyt857, line 234: This is also probably incorrect now.

It would be better to revert these for now, and make a separate PR just for them on their own so we can more easily discuss the differences.

sim/core/spell_effect.go
```diff
@@ -290,27 +290,22 @@ func (spellEffect *SpellEffect) applyAttackerModifiers(sim *Simulation, spell *S
 			spellEffect.BonusArmorPenRating += attacker.PseudoStats.BonusOHArmorPenRating
 		}
 	} else if spell.SpellSchool.Matches(SpellSchoolArcane) {
-		spellEffect.Damage *= attacker.PseudoStats.ArcaneDamageDealtMultiplier
+		spellEffect.Damage *= attacker.PseudoStats.ArcaneDamageDealtMultiplier * attacker.PseudoStats.MagicDamageDealtMultiplier
```
jimmyt857, line 293: This new pseudostat isn't necessary; just update all 5 schools. `MagicDamageDealtMultiplier` is only being used in 1 place, and it's during initialization where performance doesn't matter. This is adding another multiplication on the hot path.

sim/core/buffs.go
rosenrusinov: > I needed to change it because it now fits the buff description

It does not fit the buff description:
image

Its previous implementation is the correct one imo

sim/warlock/spell_damage_helper_function.go
```diff
@@ -0,0 +1,108 @@
+package warlock
+
+import (
+	"github.com/wowsims/wotlk/sim/core"
+	"github.com/wowsims/wotlk/sim/core/proto"
+)
+
+func (warlock *Warlock) spellDamageMultiplierHelper(sim *core.Simulation, spell *core.Spell, spellEffect *core.SpellEffect) float64 {
```
jimmyt857, line 8: This function should not be called from OnInit(). Almost all of this logic is static and should be precomputed. The only part that is not static is the execute multiplier, which can be split off into its own effect.

I think a better way to go would be to have a `func (warlock *Warlock) sharedAdditiveMultipliers() float64` function which handles ONLY the generic additive effects, and a  `func (warlock *Warlock) sharedMultiplicativeMultipliers() float64` which handles ONLY the generic multiplicative effects. Then in each of your spells, do something like this:

`DamageMultuplier: (warlock.sharedAdditiveMultipliers() + 0.05*float64(warlock.Talents.ImprovedShadowbolt)) * warlock.sharedMultiplicativeMultipliers() * xxx,`

This will clean up your code a lot and also make it a lot faster.

sim/warlock/talents.go
```diff
@@ -10,29 +10,28 @@ import (
...
 	// Demonic Embrace
-	if warlock.Talents.DemonicEmbrace > 0 {
-		warlock.AddStatDependency(stats.Stamina, stats.Stamina, 1.01+(float64(warlock.Talents.DemonicEmbrace)*0.03))
-	}
+	warlock.AddStatDependency(stats.Stamina, stats.Stamina, 1.01+(float64(warlock.Talents.DemonicEmbrace)*0.03))
 
-	// Suppression
-	warlock.AddStat(stats.SpellHit, float64(warlock.Talents.Suppression)*core.SpellHitRatingPerHitChance)
+	// Molten Skin
+	warlock.PseudoStats.DamageTakenMultiplier /= 1 + 0.02*float64(warlock.Talents.MoltenSkin)
```
jimmyt857, line 16: Most other talents like this are implemented as `*= 1 - 0.02*float64(warlock.Talents.MoltenSkin)`. Are you sure this is different?

sim/warlock/talents.go
```diff
@@ -118,20 +104,22 @@ func (warlock *Warlock) ApplyTalents() {
...
-		Duration: time.Second * 40,
-		Priority: float64(warlock.GetStat(stats.Spirit)),
-		OnGain: func(aura *core.Aura, sim *core.Simulation) {
-			warlock.AddStatDynamic(sim, stats.SpellPower, 0.2*float64(warlock.GetStat(stats.Spirit)))
-		},
-		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
-			warlock.AddStatDynamic(sim, stats.SpellPower, -0.2*float64(warlock.GetStat(stats.Spirit)))
-		},
-	})
+	warlock.GlyphOfLifeTapAura = warlock.NewTemporaryStatsAura(
```
jimmyt857, line 117: Check out the new `AddStatDependencyDynamic` for effects like this.

sim/core/buffs.go
Glandaaalf: So none of these buffs change haste? My bad if that's the case, I will revert these changes. I thought the spell tooltip was correct.

sim/warlock/talents.go
```diff
@@ -236,19 +212,19 @@ func (warlock *Warlock) setupPyroclasm() {
 }
 
 func (warlock *Warlock) setupEradication() {
-	hasteBonusPercent := float64(warlock.Talents.Eradication) * 6
+	castSpeedPercentBonus := float64(warlock.Talents.Eradication) * 6
```
jimmyt857, line 215: This would be easier to read if you use `0.06` here and `0.02` for eradication, and then remove the `/100` in the aura.

sim/warlock/talents.go
```diff
@@ -277,7 +253,8 @@ func (warlock *Warlock) setupShadowEmbrace() {
 		OnStacksChange: func(aura *core.Aura, sim *core.Simulation, oldStacks int32, newStacks int32) {
 			aura.Unit.PseudoStats.PeriodicShadowDamageDealtMultiplier /= 1.0 + 0.01*float64(warlock.Talents.ShadowEmbrace)*float64(oldStacks)
```
jimmyt857, line 254: The `0.01+float64(warlock.Talents.ShadowEmbrace)` can be precomputed outside this function.

sim/warlock/talents.go
```diff
@@ -320,7 +297,7 @@ func (warlock *Warlock) setupNightfall() {
 		},
 		OnPeriodicDamageDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, spellEffect *core.SpellEffect) {
 			if spell == warlock.Corruption { // TODO: also works on drain life...
-				if sim.RandomFloat("Nightfall") < 0.02*float64(warlock.Talents.Nightfall)+
+				if sim.RandomFloat("Nightfall") < 0.02*float64(warlock.Talents.Nightfall) +
```
jimmyt857, line 300: This should use a precomputed `procChance` outside the lambda for efficiency.

ui/core/_individual_sim_ui.scss
```diff
@@ -130,6 +130,24 @@
 .settings-section-container {
   position: absolute;
 }
+
+.player-row {
```
jimmyt857, line 134: Don't add css rules in core unless you're sure; this affects everyone! You can override specific things by putting these same rules in `ui/warlock/_sim.ts`

sim/core/buffs.go
Glandaaalf: Oh I see. They both change haste, but if it's through rating, it becomes inheritable?

ui/core/_individual_sim_ui.scss
Glandaaalf: I played a bit with the UI, and then changed my mind, and forgot to remove this part, whoops sorry!

sim/warlock/spell_damage_helper_function.go
Glandaaalf: Alright I will do that. Just a question though, about performance. When a function with an "if condition" is called, if the if is false, does the code still try to calculate what's inside the if or is it just ignored and does not affect performance?

PR #274, opened 2022-07-24
rename DeathKnight to Deathknight

rosenrusinov:


rosenrusinov:
Looking into the test failure and if i should update them

PR #294, opened 2022-07-25
Enhancement: Feral Spirit

johnolos:
I'm confident in this PR now. Confirmed calculations on the stats in a few cases in the sim. Output damage seems in-range with combatlogs from naxxramas.

**The important bits:**
```go
stats.Agility:     113, // pet scales 1% crit per 83.3 agility.
stats.Strength:    331, // pet scales 1 str = 2ap
stats.AttackPower: -20, // strange, but it is in line with other 1str=2ap scaling units in the game
```

And the strange thing. Wolves scale 0.31 or 0.61 and not 30% or 60% from shaman. Why and how, 🤷 
```go
stats.Stamina:     ownerStats[stats.Stamina] * 0.3,
stats.Armor:       ownerStats[stats.Armor] * 0.35,
stats.AttackPower: ownerStats[stats.AttackPower] * (core.TernaryFloat64(shaman.HasMajorGlyph(proto.ShamanMajorGlyph_GlyphOfFeralSpirit), 0.61, 0.31)),
```

### DIFF
--- a/sim/shaman/spirit_wolves.go
+++ b/sim/shaman/spirit_wolves.go
@@ -1,0 +1,120 @@
+package shaman
+
+import (
+	"math"
+	"strconv"
+	"time"
+
+	"github.com/wowsims/wotlk/sim/core"
+	"github.com/wowsims/wotlk/sim/core/proto"
+	"github.com/wowsims/wotlk/sim/core/stats"
+)
+
+type SpiritWolf struct {
+	core.Pet
+
+	shamanOwner *Shaman
+}
+
+type SpiritWolves struct {
+	SpiritWolf1 *SpiritWolf
+	SpiritWolf2 *SpiritWolf
+}
+
+func (SpiritWolves *SpiritWolves) EnableWithTimeout(sim *core.Simulation) {
+	SpiritWolves.SpiritWolf1.EnableWithTimeout(sim, SpiritWolves.SpiritWolf1, time.Second*45)
+	SpiritWolves.SpiritWolf2.EnableWithTimeout(sim, SpiritWolves.SpiritWolf1, time.Second*45)
+}
+
+func (SpiritWolves *SpiritWolves) CancelGCDTimer(sim *core.Simulation) {
+	SpiritWolves.SpiritWolf1.CancelGCDTimer(sim)
+	SpiritWolves.SpiritWolf2.CancelGCDTimer(sim)
+}
+
+var spiritWolfBaseStats = stats.Stats{
+	stats.Stamina:   361,
+	stats.Spirit:    109,
+	stats.Intellect: 65,
+	stats.Armor:     9616,
+
+	stats.Agility:     113,
+	stats.Strength:    331,
+	stats.AttackPower: -20,
+
+	// Add 1.8% because pets aren't affected by that component of crit suppression.
+	stats.MeleeCrit: (1.1515 + 1.8) * core.CritRatingPerCritChance,
+}
+
+func (shaman *Shaman) NewSpiritWolf(index int) *SpiritWolf {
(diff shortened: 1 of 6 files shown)


sim/shaman/spirit_wolves.go
```diff
@@ -0,0 +1,116 @@
...
+			BaseDamageMax:  311,
+			SwingSpeed:     1.5,
+			SwingDuration:  time.Millisecond * 1500,
+			CritMultiplier: 2,
+		},
+		AutoSwingMelee: true,
+	})
+
+	spiritWolf.AddStatDependency(stats.Strength, stats.AttackPower, 1.0+2)
+	spiritWolf.AddStatDependency(stats.Agility, stats.MeleeCrit, 1.0+(core.CritRatingPerCritChance/83.3))
```
johnolos, line 68: I read it was 1% crit  = 83.3 agi rating somewhere on the internett. This is just a guessimate.

Wumblesnarf: 83.3 agi = 1% crit is correct

sim/shaman/enhancement/rotation.go
```diff
@@ -41,6 +41,15 @@ func (rotation *AdaptiveRotation) DoAction(enh *EnhancementShaman, sim *core.Sim
...
 
+	if enh.Talents.FeralSpirit {
+		if enh.FeralSpirit.IsReady(sim) {
+			if !enh.FeralSpirit.Cast(sim, target) {
+				enh.WaitForMana(sim, enh.FeralSpirit.CurCast.Cost)
+			}
+			return
+		}
+	}
+
```
Wumblesnarf, line 52: id have to double check but im not sure if cooldowns go in the rotation or go somewhere else

sim/shaman/spirit_wolves.go
```diff
@@ -0,0 +1,116 @@
...
+func (SpiritWolves *SpiritWolves) CancelGCDTimer(sim *core.Simulation) {
+	SpiritWolves.SpiritWolf1.CancelGCDTimer(sim)
+	SpiritWolves.SpiritWolf2.CancelGCDTimer(sim)
+}
+
+// Source: https://web.archive.org/web/20201120214816/https://github.com/dalaranwow/dalaran-wow/issues/4670
+var spiritWolfBaseStats = stats.Stats{
+	stats.Agility:     113,
+	stats.Strength:    331,
+	stats.AttackPower: 836, // uncertain number. 1 STR = 2AP, but surely AGI has a contribution?. This is copy-value from dk/ghoul
```
Wumblesnarf, line 38: yeah if it uses the same scaling as the shaman it should be 1:1 for both str and agi to AP

johnolos: I've double-checked, and the stats are agi 113, str 331, attack power -20.

Spirit Wolf unbuffed will have 331 str * 2 = 662
-20 yielding 642 attack power.

This checks out with testing.

sim/shaman/enhancement/rotation.go
johnolos: You are correct, and I have fixed this now and removed spirit wolves explicitly from rotation.

It is prioritized above blood lust, and judging timeline, it correctly pops wolves before lust.

sim/shaman/spirit_wolves.go
johnolos: Incorrect. Wolves scale with 2 ap = 1 str.

They only scale 1-1 from AP inherited from the shaman owner. This line is not what implements that.

In other words, Kibler's bits will yield an additional 40 attack power for the wolves not accounting for the rest of the raid buffs.

PR #297, opened 2022-07-25
A variety of Rogue poison/ability fixes

b-r-a-n:
Fixes a bunch of small issues I've identified while trying to reconcile difference between current combat implementation and Simonize's sheet.

### DIFF
--- a/sim/core/spell_outcome.go
+++ b/sim/core/spell_outcome.go
@@ -24,6 +24,25 @@ func (unit *Unit) OutcomeFuncTick() OutcomeApplier {
 	}
 }
 
+func (unit *Unit) OutcomeFuncTickHitAndCrit(critMultiplier float64) OutcomeApplier {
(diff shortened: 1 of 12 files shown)


sim/core/spell_outcome.go
```diff
@@ -24,6 +24,40 @@ func (unit *Unit) OutcomeFuncTick() OutcomeApplier {
 	}
 }
 
+func (unit *Unit) OutcomeFuncTickHitAndCrit(critMultiplier float64) OutcomeApplier {
```
jimmyt857, line 27: These are OK, but just want to clarify; the only difference between the 'Tick' outcome and a 'Hit' outcome is that the Hit counter is not incremented. The only effect this has is that the Detailed Results panel in the UI will have a different value for # of hits and Dmg / Hit. So if you want to continue the same logic while allowing crits, these make sense.

b-r-a-n: My intent was to maintain the same logic and also allow crits. If I understand you comment correctly I think that means they are good as is.

PR #363, opened 2022-07-28
Deathknight rotation updates

rosenrusinov:
- disabled presim until it becomes needed by someone
- continued work on rotation code


sim/deathknight/scourge_strike.go
```diff
@@ -76,6 +76,8 @@ func (dk *Deathknight) registerScourgeStrikeSpell() {
 					if dk.DiseasesAreActive(spellEffect.Target) {
 						dk.LastScourgeStrikeDamage = spellEffect.Damage
 						shadowDamageSpell.Cast(sim, spellEffect.Target)
+						shadowDamageSpell.SpellMetrics[spellEffect.Target.Index].Casts--
+						shadowDamageSpell.SpellMetrics[spellEffect.Target.Index].Hits--
```
rosenrusinov, line 80: @lologarithm why was this done?

rosenrusinov: image
 it breaks the shadow damage reporting in details and i think should be reverted

PR #389, opened 2022-07-30
[WIP] Rotation analysis (Enhancement Shaman)

lantisnt:


proto/shaman.proto
```diff
@@ -194,11 +194,24 @@ enum ShamanImbue {
...
+	ConfigurableRotationType type = 1;
+	repeated int32 order = 2;
+}
+
 message ElementalShaman {
 	message Rotation {
 		ShamanTotems totems = 3;
 
-		enum RotationType {
+		enum RotationType { // todo use common same as enha
```
lologarithm, line 214: we probably dont want to share rotations with enh since they will be very different. In TBC we had rotations like `FixedCLtoLBRotation` and `CastCLOnClearcasting`.

Every spec will have their own types of rotations.  The only main request we have is that there be a 'one rotation that automatically does the best it can for a given talent/item set'. This allows us to drag and drop a "player" into the raid and not require the person making the raid to know every single classes optimal rotations/gear/etc

PR #395, opened 2022-07-30
Deathknight fixes and rota updates

rosenrusinov:
- fixed several procs to use proper flags
- fixed dnd to snapshot ap/crit but use dynamic modifiers
- fixed metagem haste proc icd and chance
- more updates on rotations (frost temp reverted to old rota)


sim/deathknight/scourge_strike.go
```diff
@@ -76,8 +76,6 @@ func (dk *Deathknight) registerScourgeStrikeSpell() {
 					if dk.DiseasesAreActive(spellEffect.Target) {
 						dk.LastScourgeStrikeDamage = spellEffect.Damage
 						shadowDamageSpell.Cast(sim, spellEffect.Target)
-						shadowDamageSpell.SpellMetrics[spellEffect.Target.Index].Casts--
```
lologarithm, line 79: so I removed this because it was making the average damage look weird... I am not sure the best way to handle this.

rosenrusinov: Yeah i've noticed but without this the shadow part is even messier. Maybe we can just split them somehow? I know that if I use rank3 spell Id for the shadow part it shows up as a separate ability in the details with proper metrics but thats just wrong in my head

PR #423, opened 2022-08-01
reworked rune cost system

lologarithm:


sim/deathknight/army_of_the_dead.go
```diff
@@ -29,6 +30,7 @@ func (dk *Deathknight) registerArmyOfTheDeadCD() {
 		return
 	}
 
+	// TODO: this might be handled already in core code.. double check (if not we might want to enable it)
```
rosenrusinov, line 33: It was not. I also taught that if a spell is channeled this will happen automatically so 👍 on adding it

sim/deathknight/bone_shield.go
```diff
@@ -44,13 +45,16 @@ func (dk *Deathknight) registerBoneShieldSpell() {
 		},
 	})
 
+	baseCost := float64(core.NewRuneCost(10, 0, 0, 1, 0))
```
rosenrusinov, line 48: Bone shield does not cost 10 runic power or I'm not understanding this. isnt it supposed to be -10 because it gives rp

sim/deathknight/deathknight.go
```diff
@@ -450,7 +448,7 @@ func (dk *Deathknight) CanCast(sim *core.Simulation, spell *core.Spell) bool {
 	case dk.BloodStrike:
 		return dk.CanBloodStrike(sim)
 	case dk.BloodBoil:
-		return dk.CanBloodBoil(sim)
+		return dk.BloodBoil.IsReady(sim)
```
rosenrusinov, line 451: CanBloodBoil also checks for available cost, does IsReady check for that? I taught it does only cd checks

sim/deathknight/howling_blast.go
```diff
@@ -12,25 +13,33 @@ func (dk *Deathknight) registerHowlingBlastSpell() {
 	}
 
 	rpBonus := 2.5 * float64(dk.Talents.ChillOfTheGrave)
+	baseCost := float64(core.NewRuneCost(15, 0, 1, 1, 0))
```
rosenrusinov, line 16: Wrong Cost, 0, 1, 1 should be correct

rosenrusinov: oh wait thats a death rune at the end, ignore this

PR #442, opened 2022-08-03
Warlock Demonic Pact aura and totem of wrath aura

Glandaaalf:


sim/core/buffs.go
```diff
@@ -868,3 +866,72 @@ func ReplenishmentAura(character *Character, actionID ActionID) *Aura {
 		},
 	})
 }
+
+func DemonicPactAura(character *Character, spellPowerBonus float64) *Aura {
+
+	return character.GetOrRegisterAura(Aura{
+		Label:    "Demonic Pact",
+		Tag:      "SpellPowerBuff",
```
rosenrusinov, line 874: you should add the tag as a const and use that in all the auras instead of a magic string

PR #447, opened 2022-08-03
Merge dk rune resources

### LINKED ISSUE #448: Merge deathknight runes
Currently the deathknight resources take up a large amount of space on the timeline.

To mitigate this we will merge all rune resources together under a generic category and make a new visualization, so that rune gains/expenditure can be shown in a single resource line.

Kruhlmann:
Resolves #448 

This PR also includes a test setup for the ui via jest.


ui/core/resources/merger.ts
```diff
@@ -0,0 +1,23 @@
...
+export class ResourceMerger {
+	protected readonly targets: ResourceType[];
+	protected readonly aggregate_key: ResourceType;
+
+	public constructor(targets: ResourceType[], aggregate_key: ResourceType) {
+		this.targets = targets;
+		this.aggregate_key = aggregate_key;
+	}
+
+	public mergeResources(resourceEvents: Record<ResourceType, ResourceChangedLogGroup[]>): Record<ResourceType, ResourceChangedLogGroup[]> {
```
jimmyt857, line 13: Seems like this should just be a static function with 3 parameters (adding `targets` and `aggregate_key`? This object is never used beyond a single call to the function.

jimmyt857: Actually, another alternative if you want to stick with the OO pattern: add a static `RUNE_MERGER` instance of this class which only gets created once.

Kruhlmann: The idea behind making this a generic class is that you could do something like:

```ts
const health_mana_merger = new new ResourceMerger([ResourceType.ResourceTypeHealth, ResourceType.ResourceTypeMana], ResourceType.ResourceTypeHealth)
health_mana_merger.mergeResources(some_input)
```

Which enables you to do this for any set of resources, not only runes.

Kruhlmann:
Closing in favor of https://github.com/wowsims/wotlk/pull/1674

PR #474, opened 2022-08-04
Refactor rotation into Hunter-like

lantisnt:


sim/shaman/enhancement/rotation.go
```diff
@@ -78,46 +54,38 @@ func (rotation *AdaptiveRotation) DoAction(enh *EnhancementShaman, sim *core.Sim
...
+		return
 	}
 
-	//enh.LightningShield.Cast(sim, nil) // if nothing else, refresh lightning shield
-	enh.DoNothing()
-	return
+	if !spell.Cast(sim, target) {
+		enh.WaitForMana(sim, spell.CurCast.Cost)
+	}
 }
```
johnolos, line 89: Looks alot better, but one more thing I can think of can be done before merging.

I _really_ do not like the `isReady()`-check within the `chooseSpell`. I haven't checked, maybe you have. Is it possible to move it to `DoAction`?

I recognize something along the lines of, but I am not sure if that changes the landscape of things.

```go
    spell := rotation.chooseSpell(enh, sim, target)
	if spell.IsReady() && !spell.Cast(sim, target) {
		enh.WaitForMana(sim, spell.CurCast.Cost)
	}
```

johnolos (approved):
It looks alright as it is now. Given test passes as these shouldn't yield changes to sim.

I'd like to see `IsReady` removed from `chooseSpell`, but not sure if its possible.

sim/shaman/enhancement/rotation.go
lantisnt: As per discussion on discord - If we will use `IsReady` outside of the `chooseSpell` then we need to have an option to fallback to it. This means solving this in a loop or preparing and checking chooseSpell only on Ready spells. Let's do one of those in next iteration

PR #483, opened 2022-08-05
Ankhar/mage sim

mormigil:
Tier sets 7, 8, 9
Full arcane spec talents + spells + rotation
Full fire spec talents + spells + rotation

No Mirror Image
No Advanced Cooldown timing
No multi target living bomb support
Minimal frost talents, spells or rotation


mormigil:
Mana regen had the wrong constant and was giving too much mana
Made the wotlk aoe cap function added in warlock the default and updated it's references
Added generic ICD + Stat function

PR #493, opened 2022-08-05
Naive energy tick update, Rogue rotation and crit damage fixes

b-r-a-n:
Several different updates in this change including:

1. Creating rogue specific imbues for poisons and cleaning up generic weapon imbues
2. Implementing Assassination
3. Progress on the two big combat spec issues: Rotation & Crit damage multipliers
4. UI updates including to prepare for alpha

Updating the rotation included updating the energy tick constants which is like the naive way of making the energy bar behave like it should in WotLK (at least with respect to ticks). There are potential performance considerations here but I kind of want to punt on that for now as I think it is more of a blocker for the raid sim.

The crit damage multiplier changes mainly undo a hack I had in place to match the sheet. The values I'm getting now are in line with data from beta and go back to using the generic crit damage multiplier formula.

The rotation changes are just minor tweaks to try and get casts closer to what the theoretical ideal ones that show up in Simon's sheet.

### DIFF
--- a/sim/rogue/poisons.go
+++ b/sim/rogue/poisons.go
@@ -13,18 +13,17 @@ func (rogue *Rogue) applyPoisons() {
 	rogue.applyInstantPoison()
 }
 
+var DeadlyPoisonActionID = core.ActionID{SpellID: 57973}
+
 func (rogue *Rogue) registerDeadlyPoisonSpell() {
-	actionID := core.ActionID{SpellID: 57973}
 
 	rogue.DeadlyPoison = rogue.RegisterSpell(core.SpellConfig{
-		ActionID:    actionID,
+		ActionID:    DeadlyPoisonActionID,
 		SpellSchool: core.SpellSchoolNature,
-
 		ApplyEffects: core.ApplyEffectFuncDirectDamage(core.SpellEffect{
-			ProcMask:            core.ProcMaskEmpty,
-			BonusSpellHitRating: 5 * core.SpellHitRatingPerHitChance * float64(rogue.Talents.Precision),
-			ThreatMultiplier:    1,
-			OutcomeApplier:      rogue.OutcomeFuncMagicHit(),
+			ProcMask:         core.ProcMaskEmpty,
+			ThreatMultiplier: 1,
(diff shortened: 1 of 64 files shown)


ui/rogue/presets.ts
```diff
@@ -1,138 +1,141 @@
...
 });
 
 export const DefaultConsumes = Consumes.create({
 	defaultPotion: Potions.PotionOfSpeed,
 	defaultConjured: Conjured.ConjuredRogueThistleTea,
-	battleElixir: BattleElixir.ElixirOfDeadlyStrikes,
-	food: Food.FoodFishFeast,
-	mainHandImbue: WeaponImbue.WeaponImbueRogueInstantPoison,
-	offHandImbue: WeaponImbue.WeaponImbueRogueDeadlyPoison,
+	startingConjured: Conjured.ConjuredRogueThistleTea,
```
jimmyt857, line 46: If this is the same as defaultConjured, then both this and `numStartingConjured` can be removed.

b-r-a-n: Fixed. I don't completely understand how this is used but with only the default specified, the UI option for starting conjured is set to none. So, a nice side effect of having all 3 was that the value was populated (to thistle tea in my case)

sim/rogue/poisons.go
```diff
@@ -61,22 +60,29 @@ func (rogue *Rogue) registerDeadlyPoisonSpell() {
...
 		Label:     "DeadlyPoison-" + strconv.Itoa(int(rogue.Index)),
-		ActionID:  actionID,
+		ActionID:  DeadlyPoisonActionID,
 		MaxStacks: 5,
 		Duration:  time.Second * 12,
 		OnGain: func(aura *core.Aura, sim *core.Simulation) {
-			if rogue.Talents.SavageCombat < 1 {
-				return
+			if rogue.Talents.SavageCombat > 0 {
+				savageCombatAura := core.SavageCombatAura(target, rogue.Talents.SavageCombat)
```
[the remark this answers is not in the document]
b-r-a-n: This ended up being good feedback outside of the performance considerations. It made me realize a lot of this code was also assuming single target and it forced me to fix that which will make doing the mult-target updates much easier.

PR #500, opened 2022-08-06
Add procs and ppm metrics for auras

jimmyt857:


mormigil (approved):
Cool this'll be helpful

PR #507, opened 2022-08-06
Fixed several mana bugs and blast calculation issues

mormigil:
Only mirror image, burn at end, and partial resists should be different than og mage sim now.

### DIFF
--- a/sim/mage/arcane_blast.go
+++ b/sim/mage/arcane_blast.go
@@ -1,7 +1,6 @@
 package mage
 
 import (
-	"math"
 	"time"
 
 	"github.com/wowsims/wotlk/sim/core"
@@ -28,6 +27,7 @@ func (mage *Mage) registerArcaneBlastSpell() {
 	})
 
 	actionID := core.ActionID{SpellID: 42897}
+	totalDiscount := 1 - .01*float64(mage.Talents.ArcaneFocus+mage.Talents.Precision)
 
 	mage.ArcaneBlast = mage.RegisterSpell(core.SpellConfig{
 		ActionID:    actionID,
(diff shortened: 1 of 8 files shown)


sim/mage/arcane_blast.go
```diff
@@ -28,6 +27,7 @@ func (mage *Mage) registerArcaneBlastSpell() {
 	})
 
 	actionID := core.ActionID{SpellID: 42897}
+	totalDiscount := 1 - .01*float64(mage.Talents.ArcaneFocus+mage.Talents.Precision)
```
jimmyt857, line 30: I think I added precision already, in talents.go. You might need to change that so it isn't double-dipped.

sim/mage/arcane_missiles.go
```diff
@@ -35,7 +35,7 @@ func (mage *Mage) registerArcaneMissilesSpell() {
 
 		Cast: core.CastConfig{
 			DefaultCast: core.Cast{
-				Cost: baseCost,
+				Cost: (1 - .01*float64(mage.Talents.ArcaneFocus)),
```
jimmyt857, line 38: This looks very wrong  :)

mormigil: oh lol without the base. Is there a cleaner way to do reduced cost on all arcane spells also btw? I mean I guess this works fine if I actually have the base value.

sim/mage/arcane_blast.go
mormigil: I want a better way to do this, but in essence my ModifyCast function needs to have the correct BaseManaCost after precision applies, but precision as a pseudostat.CostMultiplier only applies later. Without the discount the cost of AB is quite a bit too high because of the multiplication for the number of stacks. Also technically precision and arcane focus are supposed to be additive, but it really doesn't matter except for AB because the values aren't big enough to even notice a difference.

Anyway the way I solved the issue here is that I double count precision and then I subtract out one of the instances of precision in the cast.Cost call.

What would be best is if I had a way for precision to reduce all base costs of all the mage spells and additively stack with arcane focus, but that seemed like a lot of potential changes for some very minor mana diffferences.

sim/mage/arcane_missiles.go
jimmyt857: One neat way of doing it which I've thought about but never tried: use the unit.Spellbook array to loop over all the registered spells, check their school, and then apply it. You might need to be careful about initialization order of things though.

PR #553, opened 2022-08-08
[Protection Warrior] Update Shield Block to WOTLK version

CrowbarKZ:
- Update Shield Block spell behavior to WOTLK version (https://www.wowhead.com/wotlk/spell=2565/shield-block)
- Make it a major cooldown in the sim
- Adjust rotation option for Shield Block in the UI and protobuf to be a boolean, since it doesn't make sense with the new version to use it only to fish for revenge procs 
- Adjust rotation to prioritize using Shield Slam (needed it for testing the SB change, because it was not used at all in the rotation)
- A bit of refactoring for prot war code


sim/warrior/protection/protection_warrior.go
```diff
@@ -44,14 +44,14 @@ func NewProtectionWarrior(character core.Character, options proto.Player) *Prote
 		Options:  *warOptions.Options,
 	}
 
+	war.PseudoStats.CanBlock = war.Equip[proto.ItemSlot_ItemSlotOffHand].WeaponType == proto.WeaponType_WeaponTypeShield
```
jimmyt857, line 47: You shouldn't need to do this, it's already handled generically by `core`: https://github.com/wowsims/wotlk/blob/master/sim/core/character.go#L122

sim/warrior/shield_block.go
```diff
@@ -7,60 +7,53 @@ import (
...
-
 		Cast: core.CastConfig{
-			DefaultCast: core.Cast{
-				Cost: cost,
-			},
+			DefaultCast: core.Cast{},
 			CD: core.Cooldown{
 				Timer:    warrior.NewTimer(),
-				Duration: time.Second * 5,
+				Duration: cooldownDuration,
```
jimmyt857, line 41: Nothing wrong with this, but it would be cleaner to instead write this as:

`Duration: time.Second*60 - time.Second*10*time.Duration(warrior.Talents.ShieldMastery),`

Just so the reader doesn't need to go searching to find the value.

sim/warrior/shield_block.go
```diff
@@ -7,60 +7,53 @@ import (
...
-		warrior.StanceMatches(DefensiveStance) &&
-		warrior.CurrentRage() >= warrior.ShieldBlock.DefaultCast.Cost &&
-		warrior.ShieldBlock.IsReady(sim)
+	warrior.AddMajorCooldown(core.MajorCooldown{
+		Spell: warrior.ShieldBlock,
+		Type:  core.CooldownTypeDPS,
+		CanActivate: func(sim *core.Simulation, character *core.Character) bool {
+			return warrior.PseudoStats.CanBlock &&
+				warrior.StanceMatches(DefensiveStance) &&
+				warrior.ShieldBlock.IsReady(sim)
```
jimmyt857, line 56: The check for `IsReady()` here is unnecessary, because the `core` code for MajorCooldowns already checks for that: https://github.com/wowsims/wotlk/blob/master/sim/core/major_cooldown.go#L368

PR #621, opened 2022-08-12
Enhancement shaman

dacendo:
Locally tested and seems to work properly, do what you see fit for that small work.


Wumblesnarf:
looks good, might adjust some little things here or there but thats pretty minor

PR #672, opened 2022-08-17
lots of little enhancement shaman updates and fixes

Wumblesnarf:
also fixes the tbc JC crafted necks with the buffs to properly give both types of crit


Wumblesnarf:
having an issue with tests failing on the action, but not having any issues locally. not sure what it could be, the mage tests look like rounding errors, but the hunter tests are a bit more off

PR #688, opened 2022-08-18
Balance wotlk

Gashiraa:
Gashira#2451

### DIFF
--- a/sim/druid/talents.go
+++ b/sim/druid/talents.go
@@ -64,6 +64,7 @@ func (druid *Druid) ApplyTalents() {
 	druid.registerNaturesSwiftnessCD()
 	druid.applyPrimalFury()
 	druid.applyOmenOfClarity()
+	druid.applyEclipse()
 }
 
 func (druid *Druid) registerNaturesSwiftnessCD() {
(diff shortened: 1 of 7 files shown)


ui/raid/presets.ts
```diff
@@ -178,10 +178,10 @@ export const playerPresets: Array<PresetSpecSettings<any>> = [
 		defaultGear: {
 			[Faction.Unknown]: {},
 			[Faction.Alliance]: {
-				5: BalanceDruidPresets.P5_PRESET.gear,
+				5: BalanceDruidPresets.T7_PRESET.gear,
```
lologarithm, line 181: make this `1:` ... T7 will be phase 1 gear :)

sim/druid/talents.go
```diff
@@ -199,3 +200,70 @@ func (druid *Druid) ApplyClearcasting(sim *core.Simulation, spell *core.Spell, c
...
 }
+
+func (druid *Druid) applyEclipse() {
+	druid.SolarICD = core.Cooldown{Timer: druid.NewTimer(), Duration: 0}
+	druid.LunarICD = core.Cooldown{Timer: druid.NewTimer(), Duration: 0}
+	if druid.Talents.Eclipse == 0 {
+		return
+	}
+	// Solar
+	solarProcChance := 0.333 * float64(druid.Talents.Eclipse)
```
jimmyt857, line 211: This should probably be (1.0/3.0) instead of 0.333

PR #710, opened 2022-08-20
build: small fixes

jarveson:
- fixes 'GOROOT' with space in path
 - fixes glob.sync breaking on windows path slash
 - fix makefile color printing

Found from getting this working on windows, the `replaceAll` is probably the only 'odd' thing, otherwise nothing should cause any issues on other platforms to my knowledge. I understand if you're not interested in supporting this, but with this i was able to 'finagle' native windows with msys2 to handle makefile command

### DIFF
--- a/makefile
+++ b/makefile
@@ -48,7 +48,7 @@ $(OUT_DIR)/bundle/.dirstamp: \
 	touch $@
 
 $(OUT_DIR)/sim_worker.js: ui/worker/sim_worker.js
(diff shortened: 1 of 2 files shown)


jarveson:
awkward, quoting actually broke build. changed quote spot and changed `echo -e` to just `printf`, colors should work in terminal then locally and (hopefully) will also print in github actions correctly and without the `-e`

PR #713, opened 2022-08-20
druid/feral updates

jarveson:
Large chunk of updates for feral druid for wrath
 - Tackled effectively all talents / spells for feral
 - Merged tank/feral spells and specs a bit, fixes some crashes based on talents for previous code, also feral cat rotation meta can include 'bearweave', so those abilities will need to be available
 - Still missing glyphs / sets / rotation and probably more 'formal' verification, although tests/numbers still fall into rough expected values, added notes into spec for now
 - Indirectly gives balance druid small boost (few percent) from missing motw talent effect ( 2% stat increase )

First PR! Obviously let me know if anything is super odd or weirdly coded, i tried to roughly mimic or copy what was there or based it on other parts of the code. like some of the effects when shifting may be awkward or wrong especially, but not sure if theres a better way for now


sim/druid/berserk.go
```diff
@@ -0,0 +1,53 @@
...
+			}
+		},
+		OnExpire: func(aura *core.Aura, sim *core.Simulation) {
+			if druid.InForm(Cat) {
+				druid.PseudoStats.CostMultiplier *= 2.0
+			}
+		},
+	})
+
+	if !druid.Talents.Berserk {
```
jimmyt857, line 31: This check can be placed at the start of the function. It's better to avoid registering auras when they can't be used, because its better for performance. Also, note that many of the getter functions for auras, such as `IsActive` and `GetStacks` work fine even if the aura is `nil`.

sim/druid/enrage.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/wowsims/wotlk/sim/core"
+)
+
+func (druid *Druid) registerEnrageSpell() {
+	actionID := core.ActionID{SpellID: 5229}
+	rageMetrics := druid.NewRageMetrics(actionID)
+
+	instantRage := 20.0
+	if druid.Talents.Intensity == 1 {
```
jimmyt857, line 14: This can be written more cleanly as `instantRage := []float64{20, 24, 27, 30}[druid.Talents.Intensity]`

sim/druid/rip.go
```diff
@@ -51,34 +57,34 @@ func (druid *Druid) registerRipSpell() {
...
 	})
 }
+
+func (druid *Druid) maxRipTicks() int {
+	// TODO: This should handle sets as well
+	base := 6
+	ripGlyphBonus := core.TernaryInt(druid.HasMajorGlyph(proto.DruidMajorGlyph_GlyphOfRip), 2, 0)
+	shredGlyphBonus := core.TernaryInt(druid.HasMajorGlyph(proto.DruidMajorGlyph_GlyphOfShred), 3, 0)
+	return base + ripGlyphBonus + shredGlyphBonus
+
```
jimmyt857, line 89: blank line?

sim/druid/savage_roar.go
```diff
@@ -0,0 +1,54 @@
...
+func (druid *Druid) registerSavageRoarSpell() {
+	actionID := core.ActionID{SpellID: 52610}
+	baseCost := 25.0
+
+	druid.SavageRoarAura = druid.RegisterAura(core.Aura{
+		Label:    "Savage Roar Aura",
+		ActionID: actionID,
+		Duration: 9,
+		OnGain: func(aura *core.Aura, sim *core.Simulation) {
+			druid.PseudoStats.PhysicalDamageDealtMultiplier *= druid.getSavageRoarMultiplier()
```
jimmyt857, line 25: Instead of calling this function every time, calculate it once outside the aura callbacks and capture the constant in the closure.

sim/druid/talents.go
```diff
@@ -45,25 +57,60 @@ func (druid *Druid) ApplyTalents() {
...
 	}
 
+	if druid.Talents.MasterShapeshifter > 0 {
+		bonus := 0.02 * float64(druid.Talents.MasterShapeshifter)
+		if druid.InForm(Moonkin) {
+			druid.MultiplyStat(stats.SpellPower, 1.0+bonus)
+		} else if druid.InForm(Bear) {
+			druid.PseudoStats.DamageDealtMultiplier += bonus
+		} else if druid.InForm(Cat) {
+			druid.PseudoStats.BonusMeleeCritRating += 2 * float64(druid.Talents.MasterShapeshifter) * core.CritRatingPerCritChance
```
jimmyt857, line 110: This can just be `druid.AddStat(stats.MeleeCrit, ....)`. That pseudostat is specifically for non-ranged crit.

sim/druid/berserk.go
jarveson: interesting... did not realize thats how go works, definitely had it there cause i assumed it'd segfault otherwise ha, will change

sim/druid/talents.go
jarveson: shoot yea, somehow it didnt occur to me thered be a `stats.MeleeCrit`

jarveson:
thanks for the review, rebased and made suggested changes, no other actual changes

sim/druid/berserk.go
jimmyt857: A null dereference is still a segfault/panic in Go. However, Go allows methods to check themselves for `nil`, which we have done for these functions for convenience: https://github.com/wowsims/wotlk/blob/master/sim/core/aura.go#L131

PR #720, opened 2022-08-20
Trying to add AoE capability, but need some help :D

dbyena:
Trying to add AoE capability, but need some help :D

I've added mind sear, and the ability to do an AoE rotation, but I am unable to get mind sear to not error out. Any help is appreciated :D


dbyena:
Oh ignore the unstable affliction change. That was for science. I will remove affected by cast speed.

PR #734, opened 2022-08-21
Dk glyphs itemsets

JooperGH:
- Implemented most of the missing glyphs.
- Made AMS work for tank as well.
- Changed some presets.

### DIFF
--- a/sim/deathknight/anti_magic_shell.go
+++ b/sim/deathknight/anti_magic_shell.go
@@ -54,6 +54,7 @@ func (dk *Deathknight) registerAntiMagicShellSpell() {
 					targetDummySpell = aura.Unit.CurrentTarget.RegisterSpell(core.SpellConfig{
 						ActionID:    core.ActionID{SpellID: 49375},
 						SpellSchool: core.SpellSchoolMagic,
+						Flags:       core.SpellFlagNoOnCastComplete | core.SpellFlagNoMetrics,
 
 						Cast: core.CastConfig{},
 
--- a/sim/deathknight/anti_magic_shell.go
+++ b/sim/deathknight/anti_magic_shell.go
@@ -86,8 +86,7 @@ func (dk *Deathknight) registerAntiMagicShellSpell() {
 			dk.PseudoStats.HolyDamageTakenMultiplier *= spellDmgTakenMult
 			dk.PseudoStats.NatureDamageTakenMultiplier *= spellDmgTakenMult
 			dk.PseudoStats.ShadowDamageTakenMultiplier *= spellDmgTakenMult
-			dk.PseudoStats.DiseaseDamageTakenMultiplier *= spellDmgTakenMult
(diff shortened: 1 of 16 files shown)


sim/deathknight/anti_magic_shell.go
```diff
@@ -39,50 +39,88 @@ func (dk *Deathknight) registerAntiMagicShellSpell() {
...
+
+			dk.PseudoStats.PhysicalDamageTakenMultiplier *= physDmgTakenMult
+			dk.PseudoStats.ArcaneDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.FireDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.FrostDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.HolyDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.NatureDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.ShadowDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.DiseaseDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.PeriodicPhysicalDamageTakenMultiplier *= spellDmgTakenMult
```
jimmyt857, line 90: Should periodic physical be controlled by `physDmgTakenMult`, not `spellDmgTakenMult`?

sim/deathknight/anti_magic_shell.go
```diff
@@ -39,50 +39,88 @@ func (dk *Deathknight) registerAntiMagicShellSpell() {
...
-			sim.AddPendingAction(pa)
+
+			dk.PseudoStats.PhysicalDamageTakenMultiplier *= physDmgTakenMult
+			dk.PseudoStats.ArcaneDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.FireDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.FrostDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.HolyDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.NatureDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.ShadowDamageTakenMultiplier *= spellDmgTakenMult
+			dk.PseudoStats.DiseaseDamageTakenMultiplier *= spellDmgTakenMult
```
jimmyt857, line 89: This might be double-counting, aren't diseases also spells?

sim/deathknight/anti_magic_shell.go
JooperGH: Yes, thank you

sim/deathknight/anti_magic_shell.go
JooperGH: Probably, I've removed it for now as that makes sense, when we test the tank stuff I'll verify it

PR #737, opened 2022-08-21
add revitalize buff

JooperGH:
- Rain's implementation of revitalize


rosenrusinov:
Will update the uptime with some temporary solution before merging it as this is way too unrealistic atm

PR #740, opened 2022-08-21
Added glyphs to preset talents

ssabbyccatt:
May need to change these glyphs down the road once we know what is good, but these should be fine to start with.


Gashiraa (commented):
You're missing some imports here, on line 8 you should include : `DruidMajorGlyph, DruidMinorGlyph` then your `Glyphs.create()` should go fine

PR #749, opened 2022-08-22
SpellHaste and SpellCrit raid buffs get displayed in stats, moonkin g…

Gashiraa:
…ains those buffs from spec

### DIFF
--- a/sim/core/buffs.go
+++ b/sim/core/buffs.go
@@ -55,7 +55,9 @@ func applyBuffEffects(agent Agent, raidBuffs proto.RaidBuffs, partyBuffs proto.P
 	if raidBuffs.MoonkinAura > 0 || raidBuffs.ElementalOath {
 		character.AddStat(stats.SpellCrit, 5*CritRatingPerCritChance)
 	}
+
 	if raidBuffs.MoonkinAura == proto.TristateEffect_TristateEffectImproved || raidBuffs.SwiftRetribution {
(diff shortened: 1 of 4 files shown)


sim/core/buffs.go
```diff
@@ -55,10 +55,13 @@ func applyBuffEffects(agent Agent, raidBuffs proto.RaidBuffs, partyBuffs proto.P
...
+
 	if raidBuffs.MoonkinAura == proto.TristateEffect_TristateEffectImproved || raidBuffs.SwiftRetribution {
-		character.PseudoStats.CastSpeedMultiplier *= 1.03
-		character.PseudoStats.MeleeSpeedMultiplier *= 1.03
-		character.PseudoStats.RangedSpeedMultiplier *= 1.03
+		// For now, we assume Improved Moonkin Form is maxed-out
+		character.AddStats(stats.Stats{
+			stats.SpellHaste: 3 * HasteRatingPerHastePercent,
+			stats.MeleeHaste: 3 * HasteRatingPerHastePercent,
+		})
```
rosenrusinov, line 64: Is moonkin aura supposed to give haste rating? Attack & Cast speed multipliers and haste ratings are not the same so I think this change is actually wrong. Also changing functionality just to make something appear differently in the UI is very wrong in concept

rosenrusinov (changes requested):

PR #757, opened 2022-08-22
Haunt Logic Fixed, uptime increased.

ugurkc:
Previously, sim used to commit to spells that rendered it unable to cast haunt before it fell off, and DoT's/Drain Soul ticked sometimes without haunts. Now, haunt is applied much more conservatively, and almost every spell has a check on it that casts haunt instead if cast time of the queued spell+cast time of haunt+haunts time travelled is lower than haunt's duration.

Additionally some renames to variables in the warlock sim to make them easier to understand.

### DIFF
--- a/sim/warlock/rotations.go
+++ b/sim/warlock/rotations.go
@@ -87,26 +87,26 @@ func (warlock *Warlock) tryUseGCD(sim *core.Simulation) {
 			nextBigCD = cdReadyAt
 		}
 	}
-	allCDs := []time.Duration{
+	rotationalLeeway := []time.Duration{
 		0,
 		0,
 		0,
 	}
-
+	hauntSBTravelTime := time.Duration(float64(warlock.DistanceFromTarget)/20) * time.Second
+	hauntCastTime := warlock.ApplyCastSpeed(warlock.Haunt.DefaultCast.CastTime)
+	UACastTime := warlock.ApplyCastSpeed(warlock.UnstableAff.DefaultCast.CastTime)
+	//SBCastTime := warlock.ApplyCastSpeed(warlock.ShadowBolt.DefaultCast.CastTime)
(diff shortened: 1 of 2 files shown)


sim/warlock/rotations.go
```diff
@@ -414,7 +428,15 @@ func (warlock *Warlock) tryUseGCD(sim *core.Simulation) {
 				warlock.WaitUntil(sim, nextCD)
 				return
 			} else {
-				spell = filler
+				if warlock.Talents.Haunt &&
+					warlock.Haunt.CD.IsReady(sim) &&
+					rotationalLeeway[0]-fillerCastTime < 0 {
```
Glandaaalf, line 433: Isn't there a possibility that there is a dps increase even if the filler finishes right after `rotationalLeeway[0] == 0`? We should check if that `< 0` could be changed to something positive although small and give a dps increase. (example check lines `		if nextBigCD - sim.CurrentTime > 0 && nextBigCD - sim.CurrentTime < fillerCastTime/10 {`).

sim/warlock/rotations.go
```diff
@@ -87,26 +87,26 @@ func (warlock *Warlock) tryUseGCD(sim *core.Simulation) {
 			nextBigCD = cdReadyAt
 		}
 	}
-	allCDs := []time.Duration{
+	rotationalLeeway := []time.Duration{
```
Glandaaalf, line 90: I do not think it's a good idea to rename this right now, even though the original variable name does not match perfectly.  This gives much more "diffs"  in your commit.
Anyways, right now I'm also reforging the rotation logic, so this variable will be embedded in a new variable.

ugurkc: I'll revert this change in my branch with the fix and update accordingly.

PR #764, opened 2022-08-23
feral: 'mono-cat' rotation implementation

jarveson:
- adds in missing feral idols and tier sets
- updated base stats + other small fixes
- adds in p1 gear/talent/glyph presets
- adds in simple 'mono-cat' rotation


rosenrusinov (commented):
Damage multipliers in Wow are usually multiplicative while you have added all as additive. Are you sure this is the correct behaviour?

jarveson:
thanks for that! updated the lacerate / swipe damagemodifier calculations

sim/druid/lacerate.go
```diff
@@ -44,13 +50,13 @@ func (druid *Druid) registerLacerateSpell() {
...
 			ProcMask:         core.ProcMaskMeleeMHSpecial,
-			DamageMultiplier: 1,
+			DamageMultiplier: lbdm * dwdm,
 			ThreatMultiplier: 0.5,
 			FlatThreatBonus:  267,
 
 			BaseDamage: core.BaseDamageConfig{
 				Calculator: func(sim *core.Simulation, hitEffect *core.SpellEffect, spell *core.Spell) float64 {
-					damage := 88 + 0.01*hitEffect.MeleeAttackPower(spell.Unit)
+					damage := 88 + initialBonus + 0.01*hitEffect.MeleeAttackPower(spell.Unit)
```
rosenrusinov, line 59: This wont include attack power bonuses on the target. You can check the dk spells or  or the base attack power calculator for how to include them

jarveson: sure, updated the dot tick / initial damage

rosenrusinov: You are not multiplying the bonus ap with the 0.1 modifier as multiplication is done before addition

rosenrusinov: If feral uses custom ap damage formulas i suggest you wrap that into a helper function where you pass the multiplier and it returns you the (ap+target_ap)*multiplier

jarveson: check newer commit, ur commenting on old code still?

rosenrusinov: Looks good now yeah

jarveson:
thanks for review, i went through and updated other druid spells that were using attack power to account for `hitEffect.MeleeAttackPowerOnTarget()` or to use the base `core.BaseDamageFuncMelee` when applicable as well a few other small things

sim/druid/druid.go
```diff
@@ -242,30 +241,28 @@ func New(char core.Character, form DruidForm, selfBuffs SelfBuffs, talents proto
...
-		stats.Dodge:       -1.87 * core.DodgeRatingPerDodgeChance,
+		stats.Health:    6892, // 8227 health shown on naked character (would include tauren bonus)
+		stats.Strength:  94,
+		stats.Agility:   78,
+		stats.Stamina:   99,
+		stats.Intellect: 139,
+		stats.Spirit:    161,
+		stats.Mana:      3496,                                // 5301 mana shown on naked character
+		stats.SpellCrit: 1.85 * core.CritRatingPerCritChance, // Class-specific constant
+		stats.MeleeCrit: 7.48 * core.CritRatingPerCritChance, // 8.41% chance to crit shown on naked character screen
```
rosenrusinov, line 252: Naked character would also include the crit from base agility. Seenig how you have a slightly lower value i would guess you already account for that but asking just in case

sim/druid/ferocious_bite.go
```diff
@@ -48,7 +50,8 @@ func (druid *Druid) registerFerociousBiteSpell() {
 		},
 
 		ApplyEffects: core.ApplyEffectFuncDirectDamage(core.SpellEffect{
-			ProcMask: core.ProcMaskMeleeMHSpecial,
+			ProcMask:        core.ProcMaskMeleeMHSpecial,
+			BonusCritRating: t9bonus,
 			DamageMultiplier: 1 +
 				0.03*float64(druid.Talents.FeralAggression) +
 				core.TernaryFloat64(druid.HasSetBonus(ItemSetThunderheartHarness, 4), 0.15, 0),
```
rosenrusinov, line 57: Additive damage multipliers. Usually Wow uses multiplicative, so just making sure this is correct

sim/druid/rake.go
```diff
@@ -16,10 +16,12 @@ func (druid *Druid) registerRakeSpell() {
 
 	mangleAura := core.MangleAura(druid.CurrentTarget)
 
+	t9bonus := core.TernaryInt(druid.HasT9FeralSetBonus(2), 1, 0)
+
 	druid.Rake = druid.RegisterSpell(core.SpellConfig{
 		ActionID:    actionID,
 		SpellSchool: core.SpellSchoolPhysical,
-		Flags:       core.SpellFlagMeleeMetrics,
+		Flags:       core.SpellFlagMeleeMetrics | core.SpellFlagIgnoreResists,
```
rosenrusinov, line 24: I dont think physical spells can be resisted so that flag is probably not needed

sim/druid/savage_roar.go
```diff
@@ -47,8 +48,8 @@ func (druid *Druid) registerSavageRoarSpell() {
 
 		ApplyEffects: func(sim *core.Simulation, _ *core.Unit, spell *core.Spell) {
 			druid.SpendComboPoints(sim, spell.ComboPointMetrics())
-			druid.TigersFuryAura.Duration = time.Duration(float64(time.Second) * (9.0 + (5.0)*float64(druid.ComboPoints())))
-			druid.TigersFuryAura.Activate(sim)
+			druid.SavageRoarAura.Duration = durationBonus + time.Duration(float64(time.Second)*(9.0+(5.0)*float64(druid.ComboPoints())))
```
rosenrusinov, line 51: You should precompute the final duration number outside of the lambda for performance reasons (so its not calculated on every call)

rosenrusinov: I just saw it depends on current combo points so maybe precompute as much as you can and leave only the combo points calculation in here.

rosenrusinov: Also I'm noticing you're first spending combo points, then you're using `druid.ComboPoints()` wont thta always be 0 in this case?

sim/druid/druid.go
jarveson: yea these are all verified, might be a .01% swing or rounding error but as close as i can for now

sim/druid/ferocious_bite.go
jarveson: yea multiply is probly correct, will swap it

sim/druid/rake.go
jarveson: rake needs to ignore armor, which is what this flag looks to be for as well, unless im missing some other reason it would be ignoring armor

sim/druid/savage_roar.go
jarveson: shoot, yea spend needs to go after, thanks

PR #775, opened 2022-08-23
Revitalize Raid Buff

rosenrusinov:
Add revitalize as a raid buff

image


sim/core/aura.go
```diff
@@ -1003,9 +1003,11 @@ func ApplyFixedUptimeAura(aura *Aura, uptime float64, tickLength time.Duration)
 				if sim.RandomFloat("FixedAura") < uptime {
 					// Use random duration to compensate for increased chance collapsed into single tick.
 					randomDur := tickLength + time.Duration(float64(auraDuration-tickLength)*sim.RandomFloat("FixedAuraDur"))
+					oldDur := aura.Duration
```
jimmyt857, line 1006: Original duration is already saved in a local variable on line 984. Not sure if it might be more efficient to capture 1 less external variable though. W/e.

sim/core/buffs.go
```diff
@@ -755,6 +759,52 @@ func UnholyFrenzyAura(character *Character, actionTag int32) *Aura {
...
+		ActionID: hotID,
+		Duration: totalDuration,
+		OnGain: func(aura *Aura, sim *Simulation) {
+			pa := NewPeriodicAction(sim, PeriodicActionOptions{
+				Period:   tickPeriod,
+				NumTicks: ticks,
+				OnAction: func(s *Simulation) {
+					if s.RandomFloat("Revitalize Proc") < 0.15 {
+						if aura.Unit.HasManaBar() {
+							aura.Unit.AddMana(s, aura.Unit.BaseMana*0.01, manaMetrics, true)
```
jimmyt857, line 790: Are we sure this is 1% base mana and not 1% max mana? The latter seems more likely.

ui/core/components/multi_icon_picker.ts
```diff
@@ -88,10 +89,18 @@ export class MultiIconPicker<ModObject> extends Component {
 
 		if (this.currentValue) {
 			this.dropdownRootElem.classList.add('active');
-			this.currentValue.fillAndSet(this.buttonElem, false, true);
+			if (this.config.categoryId != null) {
```
jimmyt857, line 92: Instead of changing it like this, you could revert all changes to this function and instead modify the `getMaxValue` function below. Something like ```
const actionId = this.pickers.map(...);
if (this.categoryId == null) { return actionId; } else { return actionId == null ? null : categoryId; }

ui/core/individual_sim_ui.ts
```diff
@@ -583,6 +584,20 @@ export abstract class IndividualSimUI<SpecType extends Spec> extends SimUI {
 		] as Array<StatOption<IconInputConfig<Player<any>, any>>>);
 		otherBuffOptions.forEach(iconInput => makeIconInput(buffsSection, iconInput));
 
+		const revitalizeBuffOptions = this.splitRelevantOptions([
+			{ item: IconInputs.RevitalizeRejuvination, stats: [Stat.StatAttackPower, Stat.StatSpellPower] },
```
jimmyt857, line 588: If you want someone to apply to all sims, you can leave the `stats` list empty.

ui/core/components/multi_icon_picker.ts
rosenrusinov: That would make the category icon get the `active` class and show that a buff is active in this category without any being active (it was how i did it the first time)

ui/core/individual_sim_ui.ts
rosenrusinov: 4b65f50e86602d9d83103bf815d75845ceaa7329

sim/core/buffs.go
rosenrusinov: 4b65f50e86602d9d83103bf815d75845ceaa7329

sim/core/aura.go
rosenrusinov: 4b65f50e86602d9d83103bf815d75845ceaa7329

PR #792, opened 2022-08-24
Enh presets RaidBuffs/Debuffs/Rotation settings

Horatio27:
Updated Enh Preset values. 

Also fixed a issue with TotemOfWraith presets, setBuffs and setDebuffs. can do this in a separate PR if needed.


ui/enhancement_shaman/presets.ts
```diff
@@ -46,6 +50,11 @@ export const DefaultRotation = EnhancementShamanRotation.create({
 		fire: FireTotem.MagmaTotem,
 		water: WaterTotem.ManaSpringTotem,
 	}),
+  maelstromweaponMinStack: 3,
+  lightningboltWeave: true,
+  lavaburstWeave: false,
```
Horatio27, line 55: this might not be needed, by default it's false if left out.

Wumblesnarf: imo it doesnt hurt to have it, just for clarity reasons

PR #817, opened 2022-08-24
Code readability simplification

ugurkc:
The entire script is commented very extensively, waiting for other Warlock dev's reviews and contributions.


sim/warlock/rotations.go
```diff
@@ -1,3 +1,16 @@
+/* README(2022-08):
+Main Author: Glandalf (Discord : Glandalf#0679)
+Co-Author's: Linelo aka "The good bad guy" (Discord: Linelo#3958) , Pötiküvi(Discord: Pötiküvi#7506)
```
Glandaaalf, line 3: You should also cite Ketesh who also contributed :)

sim/warlock/rotations.go
```diff
@@ -159,28 +228,44 @@ func (warlock *Warlock) defineRotation() {
...
-	var spell *core.Spell
-	var filler *core.Spell
-	var target = warlock.CurrentTarget
-	mainSpell := warlock.Rotation.PrimarySpell
-	curse := warlock.Rotation.Curse
-	dotLag := time.Duration(10 * time.Millisecond)
+	var spell *core.Spell                          //the variable we'll be returning to the sim as our final decision
+	var filler *core.Spell                         //the filler spell we'll store, we will cast this whenever we have all our priorities in check
+	var target = warlock.CurrentTarget             //our current target
+	mainSpell := warlock.Rotation.PrimarySpell     // our primary spell
```
Glandaaalf, line 266: Well I did not change its name from the TBC sim as SB & Incinerate were primary spells, but they are basically filler spells now rather than primary spells.

Glandaaalf (approved):
Great stuff, will definitely make things clearer for the future

sim/warlock/rotations.go
ugurkc: I believe "filler" and "main" spells being used interchangeably is not that big of a problem because both are correct definitions.  An afflictions main source of damage is SB for example, yet it's not a priority. All the while you cast it at the bottom of the cast hierarchy.

sim/warlock/rotations.go
ugurkc: Done! My bad!

ugurkc:
Thanks for the reviews, the changes are pushed through. For all the code to be written in rotations.go, the same line of explanations should follow to keep the sourcecode easily readable.

PR #819, opened 2022-08-24
Warlock preset changes reverted

ugurkc:


ugurkc:
[Why do we use 'Presets', instead of BiS sets?](https://discord.com/channels/891730968493305867/975191328365355068/1012033698117456003)


Reverted due to Pokpoong's warning.

PR #857, opened 2022-08-26
Shaman Fire Elemental 

Horatio27:
Working on implementing the Fire Elemental Totem.


sim/shaman/fire_elemental_pet.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (fireElemental *FireElemental) GetPet() *core.Pet {
+	return &fireElemental.Pet
+}
+
+func (fireElemental *FireElemental) Initialize() {
+	fireElemental.registerFireBlast()
+	fireElemental.registerFireNova()
+	fireElemental.registerFireShieldAura()
```
Horatio27, line 58: was playing around with a Aura, should be rename to Dot

sim/shaman/fire_elemental_pet.go
```diff
@@ -0,0 +1,147 @@
...
+
+var fireElementalPetBaseStats = stats.Stats{
+	stats.Mana:        1789,
+	stats.Health:      994,
+	stats.Intellect:   147,
+	stats.Stamina:     327,
+	stats.SpellPower:  995,  //Estimated
+	stats.AttackPower: 1369, //Estimated
+
+	// TODO : No idea what his crit is at, he does not seem to gain any crit from owner.
```
raethkcj, line 120: Added a row in my spreadsheet for Fire Shield crit testing; he seems to gain spell crit from *his own* intellect at a rate of about 1% crit per 212 intellect, and has a base spell crit of 2.61%.

I don't know whether the sim adds spell crit from int here, or if this is just the base and the scaling is added elsewhere.

He likely gains melee crit from his own agility as well, but there are only a few ways to increase it; Blessing of Kings and Scroll of Agility. Intellect is more relevant as he gains it directly from the player.

PR #901, opened 2022-08-27
smourne choas proc on any melee

lologarithm:


sim/common/wotlk/shadowmourne.go
```diff
@@ -49,21 +49,22 @@ func init() {
 					return
 				}
 
+				if stackingAura.GetStacks() == 10 {
+					stackingAura.Deactivate(sim)
+					tempStrProc.Activate(sim)
+					choasBaneSpell.Cast(sim, player.CurrentTarget)
+				}
```
rosenrusinov, line 56: return? Also wont this currently start proccing the new stacks before the big str proc is faded? Should that be possible?

lologarithm: i went back and forth on it but I am not sure how it will work. We need to test but obv can't right now. The dps difference is like 10 dps.

sim/common/wotlk/shadowmourne.go
```diff
@@ -49,21 +49,22 @@ func init() {
...
+
 				if sim.RandomFloat("shadowmourne") > drainChance {
 					return
 				}
 
 				if !stackingAura.IsActive() {
```
jimmyt857, line 62: Activate already checks for `IsActive`, so you don't need this check. Also, letting `Activate` do the check is better so it records the aura proc for metrics.

lologarithm: ah, for some reason I thought activate would reset stacks. ill fix it

PR #925, opened 2022-08-29
Aftercast

Tharre:


sim/core/aura.go
```diff
@@ -49,6 +49,7 @@ type Aura struct {
 	active                     bool
 	activeIndex                int32 // Position of this aura's index in the activeAuras array.
 	onCastCompleteIndex        int32 // Position of this aura's index in the onCastCompleteAuras array.
+	afterCastIndex             int32 // Position of this aura's index in the afterCastAuras array.
```
jimmyt857, line 52: Adding another aura callback is going to slow down everyone's code. This can be done much more easily; just use `OnSpellHitDealt` instead.

PR #933, opened 2022-08-29
Prot warrior update

JooperGH:
- Updated all of the talents in the protection tree, except for the block part of Critical Block. (Its tricky)
- Updated gear presets & talent presets.
- Implemented new skills & effects.
- Implemented some glyphs.
- Deleted old TBC talents.


sim/warrior/devastate.go
```diff
@@ -53,6 +89,9 @@ func (warrior *Warrior) registerDevastateSpell() {
 				if spellEffect.Landed() {
 					if !warrior.ExposeArmorAura.IsActive() {
 						warrior.SunderArmorDevastate.Cast(sim, spellEffect.Target)
+						if hasGlyph {
```
Paur-SW, line 92: I might be wrong, but isn't this handled in sunder_armor.go lines 36 and 57? Could be duplicate logic here.

sim/warrior/revenge.go
```diff
@@ -4,10 +4,27 @@ import (
 	"time"
 
 	"github.com/wowsims/wotlk/sim/core"
+	"github.com/wowsims/wotlk/sim/core/proto"
 	"github.com/wowsims/wotlk/sim/core/stats"
 )
 
 func (warrior *Warrior) registerRevengeSpell(cdTimer *core.Timer) {
+	actionID := core.ActionID{SpellID: 30357}
+	warrior.revengeProcAura = warrior.RegisterAura(core.Aura{
```
Paur-SW, line 13: Nice, much cleaner this way imo.

sim/warrior/devastate.go
JooperGH: Oh! You're absolutely right, thank you!

sim/warrior/shouts.go
```diff
@@ -33,16 +33,17 @@ func (warrior *Warrior) makeShoutSpellHelper(actionID core.ActionID) *core.Spell
...
-			warrior.shoutExpiresAt = sim.CurrentTime + warrior.shoutDuration
+			warrior.shoutExpiresAt = sim.CurrentTime + warrior.shoutDuration + extraDuration
 		},
 	})
 }
 
 func (warrior *Warrior) makeShoutSpell() *core.Spell {
 	if warrior.ShoutType == proto.WarriorShout_WarriorShoutBattle {
-		return warrior.makeShoutSpellHelper(core.ActionID{SpellID: 47436})
+		return warrior.makeShoutSpellHelper(core.ActionID{SpellID: 47436}, 0)
```
Paur-SW, line 43: No Glyph of Battle? Is that handled elsewhere?

PR #941, opened 2022-08-29
Protection Warrior Update & Encounter changes

JooperGH:
- Updated all of the talents in the protection tree, except for the block part of Critical Block. (Its tricky)
- Updated gear presets & talent presets.
- Implemented new skills & effects.
- Implemented some glyphs.
- Deleted old TBC talents.
- Made use of the TargetAI interface to implement the Patchwerk AI, so that we can properly simulate boss encounters.
- Currently there's some issue with picking 10vs25 man so I've disabled 10 man. Boss values are completely artificial and need research.


sim/warrior/shouts.go
```diff
@@ -33,16 +33,17 @@ func (warrior *Warrior) makeShoutSpellHelper(actionID core.ActionID) *core.Spell
...
-			warrior.shoutExpiresAt = sim.CurrentTime + warrior.shoutDuration
+			warrior.shoutExpiresAt = sim.CurrentTime + warrior.shoutDuration + extraDuration
 		},
 	})
 }
 
 func (warrior *Warrior) makeShoutSpell() *core.Spell {
 	if warrior.ShoutType == proto.WarriorShout_WarriorShoutBattle {
-		return warrior.makeShoutSpellHelper(core.ActionID{SpellID: 47436})
+		return warrior.makeShoutSpellHelper(core.ActionID{SpellID: 47436}, 0)
```
Paur-SW, line 43: Same comment from previous PR: Where is Glyph of Battle handled if not here?

Paur-SW:
I reviewed all of the Prot Warrior updates to the best of my ability and they look great! Thank you for your hard work. Good luck with critical block 😂

sim/encounters/patchwerk10_ai.go
```diff
@@ -0,0 +1,135 @@
+package encounters
```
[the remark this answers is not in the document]
JooperGH: Good point, I'll do that

sim/warrior/shouts.go
JooperGH: I don't think its handled, I'll check and add it here if not.

sim/warrior/sunder_armor.go
```diff
@@ -69,6 +72,7 @@ func (warrior *Warrior) newSunderArmorSpell(isDevastateEffect bool) *core.Spell
 			if warrior.SunderArmorAura.GetStacks() == 5 {
```
Paur-SW, line 72: Perhaps this conditional should be removed since Devastate and Sunder Armor each do their threat independently of each other now.

sim/warrior/shockwave.go
```diff
@@ -0,0 +1,60 @@
...
+	"time"
+
+	"github.com/wowsims/wotlk/sim/core"
+	"github.com/wowsims/wotlk/sim/core/stats"
+)
+
+func (warrior *Warrior) registerShockwaveSpell() {
+	cost := 15.0 - float64(warrior.Talents.FocusedRage)
+	refundAmount := cost * 0.8
+	cd := 20 * time.Second
```
Paur-SW, line 13: Might as well drop this in
`	cd := 20 - core.TernaryDuration(warrior.HasMajorGlyph(proto.WarriorMajorGlyph_GlyphOfShockwave), 3*time.Second, 0)
`

PR #973, opened 2022-09-01
Stat weights investigation

JooperGH:
- Fixed some typos in the statweight.go file, these broke TPS/DTPS eps a fair bit.
- Made changes to the spellhit / meleehit portions of the code that still assumed 9% hit cap for melee. I want some feedback on these.


sim/core/statweight.go
```diff
@@ -134,10 +134,13 @@ func CalcStatWeight(swr proto.StatWeightsRequest, statsToWeigh []stats.Stat, ref
 			panic("Stat weights error: " + errorStr)
 		}
 		dpsMetrics := simResult.RaidMetrics.Parties[0].Players[0].Dps
-		tpsMetrics := simResult.RaidMetrics.Parties[0].Players[0].Dps
+		tpsMetrics := simResult.RaidMetrics.Parties[0].Players[0].Threat
```
jimmyt857, line 137: Nice catch, might have to backport this to the TBC sim

sim/core/statweight.go
```diff
@@ -134,10 +134,13 @@ func CalcStatWeight(swr proto.StatWeightsRequest, statsToWeigh []stats.Stat, ref
...
 		dtpsMetrics := simResult.RaidMetrics.Parties[0].Players[0].Dtps
 		dpsDiff := (dpsMetrics.Avg - baselineDpsMetrics.Avg) / value
 		tpsDiff := (tpsMetrics.Avg - baselineTpsMetrics.Avg) / value
+
+		// Im unsure this is a good way to define it, a negative result here is actually the
+		// good outcome, which is contrary to dps/tps.
```
jimmyt857, line 143: This comment is a bit confusing, I don't think this calculation favors positive or negative values?

sim/core/statweight.go
```diff
@@ -163,7 +166,21 @@ func CalcStatWeight(swr proto.StatWeightsRequest, statsToWeigh []stats.Stat, ref
 		}
 	}
 
+	// Melee hit cap is 8% in WoTLK
+	melee2HHitCap := 8 * MeleeHitRatingPerHitChance
+	if swr.Debuffs != nil && swr.Debuffs.FaerieFire == proto.TristateEffect_TristateEffectImproved {
```
jimmyt857, line 171: I believe faerie fire is spell-only hit now

sim/core/statweight.go
```diff
@@ -163,7 +166,21 @@ func CalcStatWeight(swr proto.StatWeightsRequest, statsToWeigh []stats.Stat, ref
...
+		melee2HHitCap -= 3 * MeleeHitRatingPerHitChance
+	}
+
+	// Spell hit cap is 17% in WoTLK
+	spellHitCap := 17 * SpellHitRatingPerHitChance
+	if swr.Debuffs != nil && swr.Debuffs.Misery {
```
jimmyt857, line 177: Should be Misery || Imp Faerie Fire

sim/core/statweight.go
```diff
@@ -173,8 +190,20 @@ func CalcStatWeight(swr proto.StatWeightsRequest, statsToWeigh []stats.Stat, ref
 
 	for _, stat := range statsToWeigh {
 		statMod := defaultStatMod
-		if stat == stats.SpellHit || stat == stats.MeleeHit || stat == stats.Expertise {
-			statMod = 15
+		if stat == stats.SpellHit {
+			statMod = spellHitStatMod
+			if baseStats[stat] < spellHitCap && baseStats[stat]+statMod > spellHitCap {
+				statModsHigh[stat] = baseStats[stat] - spellHitCap
```
jimmyt857, line 196: Instead of `baseStats[stat] - spellHitCap` why not just `-spellHitStatMod`? Example: if we're only 1 rating below cap, the current code here would give a stat mod of -1. That's way too low and will have huge amounts of noise.

Same comment for melee hit.

sim/core/statweight.go
```diff
@@ -268,7 +279,8 @@ func CalcStatWeight(swr proto.StatWeightsRequest, statsToWeigh []stats.Stat, ref
 		result.Tps.EpValuesStdev[stat] = result.Tps.WeightsStdev[stat] / math.Abs(result.Tps.Weights[referenceStat])
 		if result.Dtps.Weights[DTPSReferenceStat] != 0 {
 			result.Dtps.EpValues[stat] = result.Dtps.Weights[stat] / result.Dtps.Weights[DTPSReferenceStat]
-			result.Dtps.EpValuesStdev[stat] = result.Dtps.WeightsStdev[stat] / math.Abs(result.Dps.Weights[DTPSReferenceStat])
+			// Dtps not dps
```
jimmyt857, line 282: Not sure this comment is needed, makes more sense as a GH comment.

sim/core/statweight.go
```diff
@@ -301,6 +313,7 @@ func CalcStatWeight(swr proto.StatWeightsRequest, statsToWeigh []stats.Stat, ref
 	return result
 }
 
+// TODO: Get rid of this?
```
jimmyt857, line 316: This was once used in an attempt to properly calculate standard deviations. The current code uses much simpler logic but I don't think either method is fully correct.

Could probably delete this if you want, we can always recover it from the TBC code if we want it back.

sim/core/statweight.go
JooperGH: The idea is that while a dpsDiff > 0 is a thing that should result in a higher weight, the opposite is the case for dtps, as dtpsDiff < 0 is what should result in a higher weight.

sim/core/statweight.go
JooperGH: You are correct, I'm gonna fix this

sim/core/statweight.go
JooperGH: Yep.

sim/core/statweight.go
JooperGH: Good point, the idea was simply to keep the code from going over the cap in its calculations, but that can indeed result in small mods that lead to high deviation. I'll think of something smarter

sim/core/statweight.go
JooperGH: This was just me commenting when I was fixing the typo, I'll remove

sim/core/statweight.go
JooperGH: I'll clean it up, and I agree, I don't think the current way to determine deviations is correct, but it works for now. I don't particularly find EPs/weights attractive anyway

PR #974, opened 2022-09-01
Wrath and Starfire spells refactor

Gashiraa:


sim/druid/starfire.go
```diff
@@ -26,45 +26,24 @@ func (druid *Druid) registerStarfireSpell() {
...
-			// T9-4P
-			if druid.SetBonuses.balance_t9_4 {
-				spellEffect.DamageMultiplier *= 1.04
-			}
-			// Nature's Majesty
-			spellEffect.BonusCritRating += druid.TalentsBonuses.naturesMajestyBonusCrit
-		},
+
+		BonusCritRating: druid.TalentsBonuses.naturesMajestyBonusCrit +
+			core.TernaryFloat64(druid.MoonfireDot.IsActive(), core.CritRatingPerCritChance*float64(druid.Talents.ImprovedInsectSwarm), 0) +
```
jimmyt857, line 38: This only works for static effects. Dynamic effects (like dots being active) need to change the `BonusCritRating` field on the SPELL using their `OnGain` / `OnExpire` functions.

PR #975, opened 2022-09-01
Raise ilvl threshold and remove a lot of set bonus implementations

jimmyt857:


jimmyt857:
Note that we currently don't throw errors for single-item effects for items whose IDs don't exist. I left all that code in for now because I'm sure there will be a couple of them we decide to keep.

PR #978, opened 2022-09-03
mage: fix arcane blast mana cost

JanTvrdik:
The current calculation does not match observed mana cost on live. With no T5 set and both precision and arcane focus talents 3/3, the observed mana costs are 147 / 404 / 662 / 919 / 1176.


mormigil:
Is the mana cost different in the log than on live? There is another reduction beyond just this calculation where the total cost is lowered through precision so this formula alone won't spit out the final AB cost.

PR #1005, opened 2022-09-08
Add "domain=wotlk" prefix to wowhead tooltip data attribute.

### LINKED ISSUE #1002: Berserking enchant tooltip
The tooltip for the Berserking weapon enchant shows "Permanently enchant a melee weapon to sometimes increase your attack power by 21, but at the cost of reduced armor." This is the retail tooltip.

PANDAxBEARS:
Fixes #1002 

Tooltips for enchants were appearing as their retail version due to the omission of the domain parameter when populating the data-wowhead attribute. Given that this sim is scoped to wotlk I've added it to all tooltip attributes which should force tooltips to appear as their WOTLK version.


jimmyt857:
Thanks for the PR!

PR #1012, opened 2022-09-10
Fix Glyph of Sinister Strike logic.

### LINKED ISSUE #1008: Glyph of Sinister Strike is not working
sim combat rog dps with glyph of sinister strike and look that it never gives 2 cp on SS crit

PANDAxBEARS:
Fixes #1008

This PR fixes two bugs with the current implementation:
1) The calculated combo points value wasn't being passed to AddComboPoints(...).
2) The extra combo point should only be added 50% of the time on SS crits, not 50% of the time on all SS hits.


sim/rogue/sinister_strike.go
```diff
@@ -48,13 +48,13 @@ func (rogue *Rogue) registerSinisterStrikeSpell() {
 			OutcomeApplier: rogue.OutcomeFuncMeleeSpecialHitAndCrit(rogue.MeleeCritMultiplier(true, true)),
 			OnSpellHitDealt: func(sim *core.Simulation, spell *core.Spell, spellEffect *core.SpellEffect) {
 				if spellEffect.Landed() {
-					points := 1
-					if rogue.HasMajorGlyph(proto.RogueMajorGlyph_GlyphOfSinisterStrike) {
+					points := int32(1)
+					if rogue.HasMajorGlyph(proto.RogueMajorGlyph_GlyphOfSinisterStrike) && spellEffect.DidCrit() {
```
jimmyt857, line 52: Could the `HasMajorGlyph` return value be cached outside the function instead? Doing it inside the callback like this is very slow.

PANDAxBEARS: Done

jimmyt857:
Thanks for the PR, and the fix! Merging this now.

PR #1039, opened 2022-09-13
Threat of Thassarian

vigo2:
[deathknight] fixed all offhand strikes triggered by Threat of Thassarian to use the correct damage formula, reducing the impact of the "flatModifier" by up to 50%, plus minor refactoring


rosenrusinov:
Fixed in #1042

PR #1058, opened 2022-09-15
core: add 'CurrentPowerBar' concept

jarveson:
- used to correctly handle revitalize and others to correct give rage/energy based on current power bar

Signed-off-by: jarves <jarveson@gmail.com>

### DIFF
--- a/sim/common/tbc/melee_items.go
+++ b/sim/common/tbc/melee_items.go
@@ -628,11 +628,13 @@ func init() {
 		const procChance = 2.7 / 60.0
 		actionID := core.ActionID{ItemID: 29996}
 
-		var resourceMetrics *core.ResourceMetrics
+		var resourceMetricsRage *core.ResourceMetrics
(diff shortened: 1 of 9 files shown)


sim/common/tbc/melee_items.go
```diff
@@ -628,11 +628,12 @@ func init() {
 		const procChance = 2.7 / 60.0
 		actionID := core.ActionID{ItemID: 29996}
 
-		var resourceMetrics *core.ResourceMetrics
+		var resourceMetricsRage *core.ResourceMetrics
+		var resourceMetricsEnergy *core.ResourceMetrics
 		if character.HasRageBar() {
-			resourceMetrics = character.NewRageMetrics(actionID)
+			resourceMetricsRage = character.NewRageMetrics(actionID)
 		} else if character.HasEnergyBar() {
```
jimmyt857, line 635: This shouldn't be an 'else' condition anymore since both rage and energy might be enabled for druids.

PR #1062, opened 2022-09-15
Some DK stuff

vigo2:
[deathknight] commented actually used SpellID and bonus damage values for ToT-abilities (which only affects Death Strike), simplified threatOfThassarianProcMasks(), and used threatOfThassarianProc() throughout

[deathknight] fixed Heart Strike using Rank 1 instead of Rank 6, and fixed Plague Strike to use the Vicious Strikes bonus crit damage 

[deathknight] refactored various crit multiplier functions into one (bonusCritMultiplier()), and replaced all default multipliers with - well - DefaultMeleeCritMultiplier(); since RuneWeapon was handled inconsistently, it just gets default multipliers now

### DIFF
--- a/sim/deathknight/heart_strike.go
+++ b/sim/deathknight/heart_strike.go
@@ -5,14 +5,11 @@ import (
 	"github.com/wowsims/wotlk/sim/core/stats"
 )
 
-var HeartStrikeActionID = core.ActionID{SpellID: 55050}
+var HeartStrikeActionID = core.ActionID{SpellID: 55262}
 
 func (dk *Deathknight) newHeartStrikeSpell(isMainTarget bool, isDrw bool, onhit func(sim *core.Simulation, spell *core.Spell, spellEffect *core.SpellEffect)) *RuneSpell {
-	bonusBaseDamage := 250.0 + dk.sigilOfTheDarkRiderBonus()
-	weaponBaseDamage := core.BaseDamageFuncMeleeWeapon(core.MainHand, true, bonusBaseDamage, true)
-	if !isMainTarget {
-		weaponBaseDamage = core.BaseDamageFuncMeleeWeapon(core.MainHand, true, bonusBaseDamage, true)
-	}
+	bonusBaseDamage := dk.sigilOfTheDarkRiderBonus()
+	weaponBaseDamage := core.BaseDamageFuncMeleeWeapon(core.MainHand, true, 736.0+bonusBaseDamage, true)
 
 	diseaseMulti := dk.dkDiseaseMultiplier(0.1)
 	weaponMulti := 0.5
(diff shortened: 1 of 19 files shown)


sim/deathknight/heart_strike.go
```diff
@@ -5,25 +5,21 @@ import (
...
 	if !isMainTarget {
 		weaponMulti = 0.25
 	}
 
-	outcomeApplier := dk.OutcomeFuncMeleeSpecialHitAndCrit(dk.critMultiplierGoGandMoM())
+	outcomeApplier := dk.OutcomeFuncMeleeSpecialHitAndCrit(dk.bonusCritMultiplier(dk.Talents.MightOfMograine))
 	if isDrw {
-		outcomeApplier = dk.RuneWeapon.OutcomeFuncMeleeSpecialHitAndCrit(
-			dk.RuneWeapon.MeleeCritMultiplier(1.0, dk.secondaryCritModifier(dk.Talents.GuileOfGorefiend > 0, dk.Talents.MightOfMograine > 0)))
+		outcomeApplier = dk.RuneWeapon.OutcomeFuncMeleeSpecialHitAndCrit(dk.RuneWeapon.DefaultMeleeCritMultiplier())
```
rosenrusinov, line 22: You have removed Might of Mograine affecting DRW strikes? Are you sure this is correct

sim/deathknight/plague_strike.go
```diff
@@ -108,7 +105,7 @@ func (dk *Deathknight) registerDrwPlagueStrikeSpell() {
 			BonusCritRating:  (dk.annihilationCritBonus() + dk.scourgebornePlateCritBonus() + dk.viciousStrikesCritChanceBonus()) * core.CritRatingPerCritChance,
 			DamageMultiplier: weaponMulti * outbreakBonus,
 			ThreatMultiplier: 1,
-			OutcomeApplier:   dk.RuneWeapon.OutcomeFuncMeleeWeaponSpecialHitAndCrit(dk.RuneWeapon.MeleeCritMultiplier(1.0, 0.0)),
+			OutcomeApplier:   dk.RuneWeapon.OutcomeFuncMeleeWeaponSpecialHitAndCrit(dk.RuneWeapon.DefaultMeleeCritMultiplier()),
```
rosenrusinov, line 108: Should this also have increased crit damage from Vicious Strikes?

rosenrusinov (approved):
Please add a comment as to why the flat damage on the OH strikes is what it is. I have left the MH * 0.5 on purpose to make it clear from where that value comes as it is optimized during compilation anyway. I would actually prefer if you revert it to those values but I guess its not that big of a deal if you add a comment mentioning it instead

sim/deathknight/heart_strike.go
vigo2: That's a case of "since RuneWeapon was handled inconsistently, it just gets default multipliers now". I'd reason it's easier to fix once someone knows which critMultipliers actually applies to DRW, and which don't.

sim/deathknight/plague_strike.go
vigo2: Dito.

vigo2:
I've added the OH strikes spellIDs, as that's where the flat damage numbers come from.

PR #1068, opened 2022-09-16
Prot Pal Updates

psiven:
Updates for Prot Pal glyphs, talents, stats, presets, and implementation of standard 969 rotation.


sim/paladin/avengers_shield.go
```diff
@@ -31,14 +31,18 @@ func (paladin *Paladin) registerAvengersShieldSpell() {
 				damage := 1100.0 + deltaDamage*sim.RandomFloat("Damage Roll")
 				damage += hitEffect.SpellPower(spell.Unit, spell) * scaling.SP
 				damage += hitEffect.MeleeAttackPower(spell.Unit) * scaling.AP
+				damage *= core.TernaryFloat64(paladin.HasMajorGlyph(41101), 2, 1)
```
rosenrusinov, line 34: This doesnt have to be a dynamic multiplier as it never changes. Cache the value of HasMajorGlyph so you can reuse it in the next call and move this multiplir to the SpellEffect.DamageMultiplier

sim/paladin/protection/rotation.go
```diff
@@ -30,65 +30,56 @@ func (prot *ProtectionPaladin) customRotation(sim *core.Simulation) {
...
-			case int32(proto.ProtectionPaladin_Rotation_HammerOfTheRighteous):
-				if prot.HammerOfTheRighteous.IsReady(sim) {
-					prot.HammerOfTheRighteous.Cast(sim, target)
-					break rotationLoop
-				}
-			}
-		}
+	//rotationLoop:
+	
+		if prot.HasMajorGlyph(41101) && prot.AvengersShield.IsReady(sim) {
```
rosenrusinov, line 35: HasMajorGlyph should be cached at initialization of the sim and then re-used in rotation code for optimisation reasons

psiven: I think this is cleanly implemented now. Thanks

sim/paladin/avengers_shield.go
```diff
@@ -9,9 +9,10 @@ import (
 
 func (paladin *Paladin) registerAvengersShieldSpell() {
 	baseCost := paladin.BaseMana * 0.26
-
 	baseModifiers := Multiplicative{}
 	baseMultiplier := baseModifiers.Get()
+	numHits := int32(1)
+        glyphedSingleTargetAS := paladin.HasMajorGlyph(41101)
```
rosenrusinov, line 15: You should use the constants that are generated from proto for glyphs instead of a magic number

sim/paladin/protection/presets.go
```diff
@@ -60,115 +70,31 @@ var FullIndividualBuffs = &proto.IndividualBuffs{
...
+    SunderArmor: true,
+    Vindication: true,
+    ThunderClap: proto.TristateEffect_TristateEffectImproved,
+    InsectSwarm: true,
 }
 
 var P1Gear = items.EquipmentSpecFromJsonString(`{"items": [
 ]}`)
 var Phase4Gear = items.EquipmentSpecFromJsonString(`{"items": [
-	{
```
jimmyt857, line 81: Deleting this gear makes sense since it's from TBC, but shouldn't it be replaced with a new set?

Removing this preset and only using the P1Gear would be fine too. In case you haven't figured it out yet, everything in this file is used *only* for the unit tests.

sim/paladin/protection/rotation.go
```diff
@@ -30,65 +30,56 @@ func (prot *ProtectionPaladin) customRotation(sim *core.Simulation) {
...
-					break rotationLoop
-				}
-			}
-		}
+	//rotationLoop:
+	
+		if isExecutePhase && prot.HammerOfWrath.IsReady(sim) {
+			// Always cast HoW if ready
+			prot.HammerOfWrath.Cast(sim, target)
+		} else
```
jimmyt857, line 38: Would you mind being consistent with the rest of our code, and keeping the `else` and `if` pieces on the same line?

sim/paladin/protection/rotation.go
```diff
@@ -30,65 +30,56 @@ func (prot *ProtectionPaladin) customRotation(sim *core.Simulation) {
...
-			case int32(proto.ProtectionPaladin_Rotation_HammerOfTheRighteous):
-				if prot.HammerOfTheRighteous.IsReady(sim) {
-					prot.HammerOfTheRighteous.Cast(sim, target)
-					break rotationLoop
-				}
-			}
-		}
+	//rotationLoop:
+	
+		if isExecutePhase && prot.HammerOfWrath.IsReady(sim) {
```
jimmyt857, line 35: I think the indentation might not be correct here. You can run `make fmt` to fix this automatically.

sim/paladin/protection/presets.go
psiven: I definitely had not figured that out... Thank you.

sim/paladin/protection/rotation.go
psiven: Done

sim/paladin/protection/rotation.go
psiven: I think I cleaned this up, but did not commit make fmt as this touched a LOT more...

rosenrusinov:
You need to sort your conflicts before we can approve this @psiven

psiven:
I believe this should resolve the AS conflict and test results.

PR #1089, opened 2022-09-18
Updated DS code to properly account for DE

dbyena:
Drain soul was not increasing dps by 12% with DE, so updated multiplier to account for DE


dbyena:
Current implementation was intended

PR #1117, opened 2022-09-21
Prot Pal Updates

psiven:
Additions:
- New Prot Pal presets for Preraid and P1 Raid, P1 default to match standard
- Reintroduced customizable rotation, options are now functional

Fixes:
- Holy Shield still had TBC threat bonus modifier
- Touched by the Light was granting 160% STR as SP instead of 60%
- Reckoning was proccing off avoided swings, also needed a ProcMask assigned due to recent changes
- Precision to Gloves enchant was also granting the Wisdom to Cloak threat reduction effect because they share an id


sim/paladin/talents.go
```diff
@@ -195,7 +195,7 @@ func (paladin *Paladin) applyReckoning() {
 				ActionID:    actionID,
 				SpellSchool: core.SpellSchoolPhysical,
 				Flags:       core.SpellFlagMeleeMetrics,
-
+				ProcMask:    core.ProcMaskMeleeMH,
```
jimmyt857, line 198: For consistency, can we put `ProcMask` after `SpellSchool` and before `Flags` like everywhere else?

ui/protection_paladin/presets.ts
```diff
@@ -74,8 +76,8 @@ export const DefaultConsumes = Consumes.create({
 	prepopPotion:  Potions.IndestructiblePotion,
 });
 
-export const P1_PRESET = {
-	name: 'P1 Preset',
+export const P0_PRESET = {
```
jimmyt857, line 79: Name this variable `PRERAID_PRESET` for consistency.

PR #1138, opened 2022-09-22
Remove school-specific spell power from stats

jimmyt857:


jimmyt857:
Note that we still support these schools, but not as stats. Instead, they are handled as +BonusSpellPower on each individual spell with the matching school type.

jimmyt857:
This PR is ~3% performance improvement, measured based on the enhance shaman sim.

PR #1226, opened 2022-10-07
Implement Avoidance Diminishing Returns

psiven:
Implement diminishing returns mechanic for avoidance (dodge/parry/miss) based on coefficients previously published by Itank; example reference here: https://github.com/Itank84/itankadin/wiki/Diminishing-Returns-on-Avoidance

- BaseDodge and BaseParry implemented as Pseudostats to handle base values and talents which are not modified by diminishing returns
- Dodge and Parry stats continue to reflect raw rating, including STR/AGI conversions as appropriate, which is true to how DR treats them as a lump sum to be diminished
- Dodge and Parry talents (Deflection, Anticipation, Feral Swiftness) moved from the latter to the former
- DK Tanks now properly have Parry enabled
- DK Spell Deflection includes Parry DR
- Moved 5% base parry from the Target AttackTable (now applies the level penalty only) to War/Pal/DK BaseParry

Did NOT touch:
- Frigid Dreadplate, which is already in the unaffected category (and possibly should not work vs bosses?)
- The left-side stats UI, which is already iffy for displaying these stats and will still need some love
- Attack table caching; this is all done live in the spell outcome calcs.

### DIFF
--- a/sim/core/avoid_dr.go
+++ b/sim/core/avoid_dr.go
@@ -22,7 +22,7 @@ const Diminish_kCm_Nondruid = (Diminish_k_Nondruid * Diminish_Cm)
 // Non-diminishing sources are added separately in spell outcome funcs
 
 
-func (unit *Unit) getDiminishedDodgeChance() (float64) {
+func (unit *Unit) GetDiminishedDodgeChance() (float64) {
 	
 	// undiminished Dodge % = D
 	// diminished Dodge % = (D * Cd)/((k*Cd) + D)
@@ -31,14 +31,14 @@ func (unit *Unit) getDiminishedDodgeChance() (float64) {
 		unit.stats[stats.Dodge]/DodgeRatingPerDodgeChance/100 +
 		unit.stats[stats.Defense]*DefenseRatingToChanceReduction
 		
-	if !unit.PseudoStats.CanParry {
-		return (dodgeChance * Diminish_Cd_Druid) / (Diminish_kCd_Druid + dodgeChance)
-	} else {
+	if unit.PseudoStats.CanParry {
 		return (dodgeChance * Diminish_Cd_Nondruid) / (Diminish_kCd_Nondruid + dodgeChance)
+	} else {
+		return (dodgeChance * Diminish_Cd_Druid) / (Diminish_kCd_Druid + dodgeChance)
 	}
 }
 
-func (unit *Unit) getDiminishedParryChance() (float64) {
+func (unit *Unit) GetDiminishedParryChance() (float64) {
 	
 	// undiminished Parry % = P
 	// diminished Parry % = (P * Cp)/((k*Cp) + P)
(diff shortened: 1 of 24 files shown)


JooperGH (approved):

sim/core/avoid_dr.go
```diff
@@ -0,0 +1,67 @@
...
+func (unit *Unit) getDiminishedDodgeChance() (float64) {
+	
+	// undiminished Dodge % = D
+	// diminished Dodge % = (D * Cd)/((k*Cd) + D)
+
+	dodgeChance := 
+		unit.stats[stats.Dodge]/DodgeRatingPerDodgeChance/100 +
+		unit.stats[stats.Defense]*DefenseRatingToChanceReduction
+		
+	if !unit.PseudoStats.CanParry {
```
jimmyt857, line 34: When otherwise irrelevant, boolean conditions like this are more readable without the negation. So just do `if unit.PseudoStats.CanParry` and flip the conditions.

Same comment for below.

sim/deathknight/talents_blood.go
```diff
@@ -97,7 +97,7 @@ func (dk *Deathknight) applySpellDeflection() {
 
 	dk.AddDynamicDamageTakenModifier(func(sim *core.Simulation, spell *core.Spell, spellEffect *core.SpellEffect) {
 		if spell.ProcMask.Matches(core.ProcMaskSpellDamage) {
-			procChance := dk.GetStat(stats.Parry) / core.ParryRatingPerParryChance
+			procChance := dk.GetStat(stats.Parry) / core.ParryRatingPerParryChance + dk.PseudoStats.BaseParry
```
jimmyt857, line 100: Should this make use of `getDiminishedParryChance`?

sim/druid/forms.go
```diff
@@ -78,7 +78,9 @@ func (druid *Druid) applyFeralShift(sim *core.Simulation, enter_form bool) {
 		}
 	}
 	druid.AddStatDynamic(sim, stats.MeleeCrit, pos*float64(druid.Talents.SharpenedClaws)*2*core.CritRatingPerCritChance)
-	druid.AddStatDynamic(sim, stats.Dodge, pos*core.DodgeRatingPerDodgeChance*2*float64(druid.Talents.FeralSwiftness))
+	//druid.AddStatDynamic(sim, stats.Dodge, pos*core.DodgeRatingPerDodgeChance*2*float64(druid.Talents.FeralSwiftness))
```
jimmyt857, line 81: Delete this old line

sim/warrior/talents.go
```diff
@@ -9,10 +9,10 @@ import (
 )
 
 func (warrior *Warrior) ApplyTalents() {
-	warrior.AddStat(stats.Parry, core.ParryRatingPerParryChance*1*float64(warrior.Talents.Deflection))
 	warrior.AddStat(stats.MeleeCrit, core.CritRatingPerCritChance*1*float64(warrior.Talents.Cruelty))
 	warrior.AddStat(stats.MeleeHit, core.MeleeHitRatingPerHitChance*1*float64(warrior.Talents.Precision))
-	warrior.AddStat(stats.Dodge, core.DodgeRatingPerDodgeChance*1*float64(warrior.Talents.Anticipation))
+	warrior.PseudoStats.BaseDodge += 0.01*float64(warrior.Talents.Anticipation)
```
jimmyt857, line 14: Let's move these lines down 1 more so that all the `warrior.AddStat` calls are kept together.

PR #1236, opened 2022-10-08
Warrior: bugfix T10 4 set proccing on non SD / BloodSurge proc

TheGroxEmpire:


sim/warrior/talents.go
```diff
@@ -591,6 +602,12 @@ func (warrior *Warrior) applySuddenDeath() {
 				if warrior.SuddenDeathAura.GetStacks() <= 1 {
 					warrior.SuddenDeathAura.SetStacks(sim, 1)
 				}
+				if sim.RandomFloat("T10 4 set") < 0.2 && Ymirjar4Set {
```
jimmyt857, line 605: The set bonus check should go first, so you don't call `RandomFloat` unless necessary. RNG is very expensive to calculate

TheGroxEmpire: Good to know, changed in [e29b9e7](https://github.com/wowsims/wotlk/pull/1236/commits/e29b9e7831c74232f03e4af8b4c3a52e066a4807)

PR #1270, opened 2022-10-11
Vigo/various updates

vigo2:
[core] SpellFlagIgnoreModifiers no longer includes SpellFlagIgnoreResists; removed SpellFlagIgnoreResists from various spells
[core] BonusPhysicalDamageTaken is now applied before all damage taken modifiers; added TargetDamageMultiplier(), the counterpart of AttackerDamageMultiplier()
[core] simplified various OutcomeApplier factories
[core] removed unused PseudoStats.DiseaseDamageDealtMultiplier
[core] added AttackTable.DamageTakenMultiplier, and used DamageDealtMultiplier as a TargetDamageMultiplier (e.g. for dot snapshotting)
[deathknight] damage multipliers now stack multiplicatively
[deathknight] scourge strike's shadow damage and wandering plague are now properly scaling from shadow damage multipliers
[druid] the feral tests were using 5/3 predatory instincts, so most results are substantially lower
[druid] brambles now affects treant damage ;)
[druid] genesis now adds its bonus additively
[hunter] fixed black arrow scaling (2.3% instead of 2% per tick)
[mage] Ignite now has partial resists
[rogue] poisons now have ProcMaskSpellDamage; Black Magic users rejoice
[rogue] fixed Backstab and Hemorrhage damage calculations; Sinister Calling affects them both multiplicatively now
[rogue] Shiv and the poison it applies now can't crit
[warlock] Chaos Bolt now has partial resists
[warrior, rogue] Sweeping Strikes and Blade Flurry use SpellFlagIgnoreAttackerModifiers now

### DIFF
--- a/sim/core/target.go
+++ b/sim/core/target.go
@@ -204,7 +204,6 @@ func (target *Target) GetMetricsProto(numIterations int32) *proto.UnitMetrics {
 }
 
 // Holds cached values for outcome/damage calculations, for a specific attacker+defender pair.
-//
 // These are updated dynamically when attacker or defender stats change.
 type AttackTable struct {
 	Attacker *Unit
@@ -228,8 +227,9 @@ type AttackTable struct {
 	PartialResistShadowThresholds Thresholds
 
 	DamageDealtMultiplier               float64
-	NatureDamageDealtMultiplier         float64
-	PeriodicShadowDamageDealtMultiplier float64
+	DamageTakenMultiplier               float64
+	NatureDamageTakenMultiplier         float64
+	PeriodicShadowDamageTakenMultiplier float64
 	HealingDealtMultiplier              float64
 }
 
@@ -239,8 +239,9 @@ func NewAttackTable(attacker *Unit, defender *Unit) *AttackTable {
 		Defender: defender,
 
 		DamageDealtMultiplier:               1,
-		NatureDamageDealtMultiplier:         1,
-		PeriodicShadowDamageDealtMultiplier: 1,
+		DamageTakenMultiplier:               1,
+		NatureDamageTakenMultiplier:         1,
+		PeriodicShadowDamageTakenMultiplier: 1,
 		HealingDealtMultiplier:              1,
 	}
 
--- a/sim/druid/insect_swarm.go
+++ b/sim/druid/insect_swarm.go
@@ -29,10 +29,10 @@ func (druid *Druid) registerInsectSwarmSpell() {
 			},
 		},
 
-		DamageMultiplier: 1 *
-			druid.TalentsBonuses.genesisMultiplier *
-			(1 + core.TernaryFloat64(druid.SetBonuses.balance_t7_2, 0.1, 0) +
-				core.TernaryFloat64(druid.HasMajorGlyph(proto.DruidMajorGlyph_GlyphOfInsectSwarm), 0.3, 0)),
+		DamageMultiplier: 1 +
+			druid.talentBonuses.genesis +
+			core.TernaryFloat64(druid.setBonuses.balance_t7_2, 0.1, 0) +
+			core.TernaryFloat64(druid.HasMajorGlyph(proto.DruidMajorGlyph_GlyphOfInsectSwarm), 0.3, 0),
 		ThreatMultiplier: 1,
 
 		ApplyEffects: core.ApplyEffectFuncDirectDamage(core.SpellEffect{
@@ -60,14 +60,14 @@ func (druid *Druid) registerInsectSwarmSpell() {
 				druid.Wrath.DamageMultiplier /= 1 + 0.01*float64(druid.Talents.ImprovedInsectSwarm)
 			},
 		}),
-		NumberOfTicks: 6 + druid.TalentsBonuses.naturesSplendorTick,
+		NumberOfTicks: 6 + druid.talentBonuses.naturesSplendor,
 		TickLength:    time.Second * 2,
 		TickEffects: core.TickFuncSnapshot(target, core.SpellEffect{
 			IsPeriodic:     true,
 			BaseDamage:     core.BaseDamageConfigMagicNoRoll(215, 0.2),
 			OutcomeApplier: druid.OutcomeFuncTick(),
 			OnPeriodicDamageDealt: func(sim *core.Simulation, spell *core.Spell, spellEffect *core.SpellEffect) {
-				if sim.RandomFloat("Elune's Wrath proc") > (1-0.08) && druid.SetBonuses.balance_t8_4 {
+				if sim.RandomFloat("Elune's Wrath proc") > (1-0.08) && druid.setBonuses.balance_t8_4 {
 					tierProc := druid.GetOrRegisterAura(core.Aura{
 						Label:    "Elune's Wrath",
 						ActionID: core.ActionID{SpellID: 64823},
--- a/sim/druid/moonfire.go
+++ b/sim/druid/moonfire.go
@@ -19,7 +19,7 @@ func (druid *Druid) registerMoonfireSpell() {
 
 	// T9-2P
 	dotOutcomeApplier := druid.OutcomeFuncTick()
-	if druid.SetBonuses.balance_t9_2 {
+	if druid.setBonuses.balance_t9_2 {
 		dotOutcomeApplier = druid.OutcomeFuncMagicHitAndCrit()
 	}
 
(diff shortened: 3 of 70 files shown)


sim/core/target.go
```diff
@@ -228,8 +227,9 @@ type AttackTable struct {
 	PartialResistShadowThresholds Thresholds
 
 	DamageDealtMultiplier               float64
-	NatureDamageDealtMultiplier         float64
-	PeriodicShadowDamageDealtMultiplier float64
+	DamageTakenMultiplier               float64
```
jimmyt857, line 230: Having `XXXTakenMultiplier` (as opposed to `XXXDealtMultiplier`) doesn't make sense on an `AttackTable`, because the table is inherently mono-directional, i.e. is only used with a particular attacker and particular defender.

sim/druid/insect_swarm.go
```diff
@@ -29,10 +29,10 @@ func (druid *Druid) registerInsectSwarmSpell() {
 			},
 		},
 
-		DamageMultiplier: 1 *
-			druid.TalentsBonuses.genesisMultiplier *
-			(1 + core.TernaryFloat64(druid.SetBonuses.balance_t7_2, 0.1, 0) +
-				core.TernaryFloat64(druid.HasMajorGlyph(proto.DruidMajorGlyph_GlyphOfInsectSwarm), 0.3, 0)),
+		DamageMultiplier: 1 +
```
jimmyt857, line 32: There is now a alternative `DamageMultiplierAdditive` on spells which can be used instead if you want. Since all of these are additive it wont make a difference, UNLESS there is some other effect which dynamically modifies either `spell.DamageMultiplier` or `spell.DamageMultiplierAdditive`.

sim/druid/moonfire.go
```diff
@@ -68,13 +67,13 @@ func (druid *Druid) registerMoonfireSpell() {
 			SpellSchool: core.SpellSchoolArcane,
 			ProcMask:    core.ProcMaskSpellDamage,
 
-			DamageMultiplierAdditive: 1 +
+			DamageMultiplier: 1 +
```
jimmyt857, line 70: Any particular reason to use the non-additive version here?

sim/hunter/black_arrow.go
```diff
@@ -60,19 +60,19 @@ func (hunter *Hunter) registerBlackArrowSpell(timer *core.Timer) {
...
 		}),
 		NumberOfTicks: 5,
 		TickLength:    time.Second * 3,
 		TickEffects: core.TickFuncSnapshot(target, core.SpellEffect{
 			IsPeriodic: true,
-
+			// scales slightly better (11.5%) than the tooltip implies (10%), but isn't affected by Hunter's Mark
 			BaseDamage: core.BuildBaseDamageConfig(func(sim *core.Simulation, spellEffect *core.SpellEffect, spell *core.Spell) float64 {
-				return 553 + 0.02*spell.RangedAttackPower(spellEffect.Target)
+				return 553 + 0.023*(spell.Unit.GetStat(stats.RangedAttackPower)+spell.Unit.PseudoStats.MobTypeAttackPower)
```
jimmyt857, line 75: You might want to mention this to Sixx in the hunter discord.

sim/hunter/serpent_sting.go
```diff
@@ -59,7 +59,7 @@ func (hunter *Hunter) registerSerpentStingSpell() {
 			Tag:      "SerpentSting",
 			ActionID: actionID,
 			OnGain: func(aura *core.Aura, sim *core.Simulation) {
-				hunter.AttackTables[aura.Unit.UnitIndex].DamageDealtMultiplier *= noxiousStingsMultiplier
+				hunter.AttackTables[aura.Unit.UnitIndex].DamageTakenMultiplier *= noxiousStingsMultiplier
```
jimmyt857, line 62: Repeating from an earlier comment, but I don't think this change should be necessary. The fact that we're retrieving the specific attack table already implies a attacker --> defender relationship, so there only needs to be 1 damage multiplier value.

sim/hunter/talents.go
```diff
@@ -262,7 +262,7 @@ func (hunter *Hunter) applyPiercingShots() {
 		ActionID:    actionID,
 		SpellSchool: core.SpellSchoolPhysical,
 		ProcMask:    core.ProcMaskEmpty,
-		Flags:       core.SpellFlagNoOnCastComplete | core.SpellFlagIgnoreModifiers,
+		Flags:       core.SpellFlagNoOnCastComplete | core.SpellFlagIgnoreModifiers | core.SpellFlagIgnoreResists,
```
jimmyt857, line 265: I believe `SpellFlagIgnoreModifiers` already includes `SpellFlagIgnoreResists`.

sim/rogue/backstab.go
```diff
@@ -31,19 +31,19 @@ func (rogue *Rogue) registerBackstabSpell() {
 		BonusCritRating: core.TernaryFloat64(rogue.HasSetBonus(ItemSetVanCleefs, 4), 5*core.CritRatingPerCritChance, 0) +
 			10*core.CritRatingPerCritChance*float64(rogue.Talents.PuncturingWounds),
 		// All of these use "Apply Aura: Modifies Damage/Healing Done", and stack additively (up to 142%).
-		DamageMultiplierAdditive: 1 +
+		DamageMultiplier: 1.5 * (1 +
```
jimmyt857, line 34: Again, why not use the additive field here?

sim/core/target.go
vigo2: The XXXDealtMultipliers are all used in applyTargetModifiers(), so they were acting like XXXTakenMultipliers to start with. I think they were introduced for Black Arrow and Noxious Sting, which both act like "Mod Damage Taken" multipliers (they apply to ticking DoTs), and have then been extended to Crusade or Improved Tracking: those, however, are "Mod Damage Done" effects, only applying to DoT snapshotting.

sim/druid/insect_swarm.go
vigo2: I'd rather remove that alternative multiplier alltogether, the only actual use case is for hunters, and they could simply handle it locally. More options = more ways of using them wrongly.

sim/druid/moonfire.go
vigo2: See above.

sim/hunter/serpent_sting.go
vigo2: See above: taken and dealt multipliers work differently, mostly wrt DoTs

sim/hunter/talents.go
vigo2: Nope. It's basically used for spells that snapshot other spells' damage (e.g. Blade Flurry, Sweeping Strikes, Necrosis), to prevent double dipping. SpellFlagIgnoreResists is now effectively a "SpellFlagBleeding", since there are no spells without (level-based) partial resists anymore.

sim/rogue/backstab.go
vigo2: Simplicity.

sim/core/target.go
jimmyt857: Ahh ok, that makes sense then, thank you.

sim/druid/insect_swarm.go
jimmyt857: Full agree that fewer options is better, but it can't be done for hunter locally without being a mess; hunters have some dynamic effects that need to stack additively with the other effects in `DamageMultiplierAdditive`.

I'm pretty sure warlocks are going to need it also, once I clean up their code.

But we can let this be for now. If you really want to remove it, it can be discussed in a separate PR.

sim/hunter/talents.go
jimmyt857: No haha, I mean this change is literally redundant: https://github.com/wowsims/wotlk/blob/8e734157beac1a64c1eb08b991d3ec39ea4f1c4e/sim/core/flags.go#L194

sim/core/target.go
vigo2: Yes, I'll add this in the next PR.

sim/hunter/talents.go
vigo2: I think it's for the bleed effect, will take a look.

sim/core/target.go
jimmyt857: This isn't the type of thing to do in the next PR; I'm just asking for a comment.

sim/hunter/talents.go
jimmyt857: There are other effects which ignore resists. A good example would be mage Ignite and hunter Piercing Shots; these ignore resistances so that they don't double-dip on the resistance calculation.

PR #1278, opened 2022-10-12
Sparse-updates

vigo2:
[common] Icebreaker is now using 4 ppm, as per testing; Icebreaker and Rune of Razorice both updated to ProcMaskSpellDamage (they both proc Black Magic)
[core] sprinkled some comments
[warrior] reduced Sweeping Strike charges to 5, as per sad reality

Testing Icebreaker (using two 2.7 speed weapons):
https://classic.warcraftlogs.com/reports/vZwNHJf8c1mr9gGd

Icebreaker into Black Magic:
Screenshot 2022-10-12 at 23 54 37

Rune of Razorice into Black Magic:
Screenshot 2022-10-12 at 23 40 18

### DIFF
--- a/sim/common/wotlk/enchant_effects.go
+++ b/sim/common/wotlk/enchant_effects.go
@@ -65,12 +65,12 @@ func init() {
 			return
 		}
 		procMask := core.GetMeleeProcMaskForHands(mh, oh)
-		ppmm := character.AutoAttacks.NewPPMManager(7.0, procMask)
+		ppmm := character.AutoAttacks.NewPPMManager(4.0, procMask)
 
 		procSpell := character.RegisterSpell(core.SpellConfig{
 			ActionID:    core.ActionID{SpellID: 44525},
 			SpellSchool: core.SpellSchoolFire,
-			ProcMask:    core.ProcMaskEmpty,
+			ProcMask:    core.ProcMaskSpellDamage,
 
 			DamageMultiplier: 1,
 			CritMultiplier:   character.DefaultSpellCritMultiplier(),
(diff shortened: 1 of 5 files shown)


TheGroxEmpire (approved):
Nice catch on the sweeping strikes charge. Warrior stuff LGTM.

sim/common/wotlk/enchant_effects.go
```diff
@@ -269,7 +269,7 @@ func init() {
 		w.BaseDamageMax += 15
 	})
 
-	core.NewItemEffect(54998, func(agent core.Agent) {
+	core.NewItemEffect(41091, func(agent core.Agent) {
```
jimmyt857, line 272: This ID needs to match the value here: https://github.com/wowsims/wotlk/blob/8e734157beac1a64c1eb08b991d3ec39ea4f1c4e/sim/core/items/all_enchants.go#L104

So this would break the effect.

Also, seems like the ID was already correct to me: https://www.wowhead.com/wotlk/spell=54998/hand-mounted-pyro-rocket

sim/core/target.go
```diff
@@ -226,8 +226,10 @@ type AttackTable struct {
 	PartialResistNatureThresholds Thresholds
 	PartialResistShadowThresholds Thresholds
 
-	DamageDealtMultiplier               float64
-	DamageTakenMultiplier               float64
+	// These damage multipliers are used if they apply to a specific attacker-defender pair only,
+	//  e.g. Stormstrike increases the caster's damage vs the debuffed mob, but noone else is affected.
```
jimmyt857, line 230: This whole `AttackTable` applies to a specific attacker-defender pair, so that part is irrelevant. Comment can say something like, `DamageDealt fields are applied as part of AttackerMultiplier(), and DamageTaken fields are applied in TargetMultiplier().`.

PR #1294, opened 2022-10-13
Testing stability

vigo2:
[core] major cooldowns were (sometimes) sorted with sort.Slice(), so go versions >= 1.18.7 (with a different implementation) had different test results

[core] majorCooldownManager.sortOne() wasn't nil-checking SharedCDs, and treated all spells without SharedCDs as sharing a CD

[druid] another instance of sort.Slice() with potential for instability 

[rogue] hand-written reverse

[many] hence, test results changed all over


vigo2:
TryUseCooldowns() should probably be dumbed down a lot, and just fully sort cooldowns between activations. I tested something like this, and it wasn't noticeably slower.

PR #1307, opened 2022-10-14
[Rogue] Fix Mutilate cost when set bonus and glyph are applied

b-r-a-n:
Net effect is that Mutilate costs 52 energy in this scenario rather than 53.

### DIFF
--- a/sim/rogue/talents.go
+++ b/sim/rogue/talents.go
@@ -72,7 +72,7 @@ func (rogue *Rogue) makeFinishingMoveEffectApplier() func(sim *core.Simulation, 
 	}
 }
 
-func (rogue *Rogue) makeCastModifier() func(*core.Simulation, *core.Spell, *core.Cast) {
+func (rogue *Rogue) makeCastModifier(staticReduction float64) func(*core.Simulation, *core.Spell, *core.Cast) {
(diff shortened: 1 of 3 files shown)


sim/rogue/talents.go
```diff
@@ -72,7 +72,7 @@ func (rogue *Rogue) makeFinishingMoveEffectApplier() func(sim *core.Simulation,
 	}
 }
 
-func (rogue *Rogue) makeCastModifier() func(*core.Simulation, *core.Spell, *core.Cast) {
+func (rogue *Rogue) makeCastModifier(staticReduction float64) func(*core.Simulation, *core.Spell, *core.Cast) {
```
vigo2, line 75: Since 4t7 (Bonescythe) is a static cost reduction, it should be used for spell costs directly. That's faster, because it essentially skips a step in the spellFunc, and simpler, because you can see the cost right where the spell is defined. And it allows to directly special case each spell (e.g. Sinister Strike's cost is _also_ reduced by 3 energy).

The cast modifier could just return nil unless 2t9 (VanCleefs) is actually equipped, simplifying the spell.castFn a little.

b-r-a-n:
Not sure if this is correct; Closing it for now.

Issue #1194, opened 2022-10-02
Crash Report 1360695815


lewie-donckers:
Link:
https://wowsims.github.io/wotlk/hunter/#eJztUr9rFEEY3ff2kpuMBEYRWQ+LyzUeQmT2lgVPkF1TXWER8B+IhaksopVaBRGUCJLOGIRITKKNaCMmRUBBMBHlwgXOXEyQqIQUKVKmUfxmt7G0iJ2PeR/zPr5fM3y6pGhoeQaDHMII74IvwRVgBxjoVSizihTX8Qw44pUedN1C9+CVSzcuX1O+YekRtDL3Vlmh5+mC2V6k6J9PGHxfo+j5DnWvubPOoLnBij8+T33ITMwxWOr4FXgifj1lsPWFThTN1Ex2UWZWQr5uQxvzYlpyJbh76zXlSM39WWdfzTm7u+js/RnXtdNh0N6gdvnvXdZbVpSihbyscGpcBvyPP3FTKRgGHMHZq/pcg1UezvzpVDpwQlN1CYvCHqESaqEvRDj50GEhqUcZVhNZjNOKpWOySZBt8Rv+BPAG/hKwBvTl/faS+qePDrtJFWM4Gdtav43iKIrDKLSRtWFUiyLxWJFx2B9bh9o3BOpxk2a6yWDhB8ufW6x+aNG2W9xBYQ/FSXjr8Cqb7MnajA6n+Tu8o+nxfNLlpJ57VpK+4QztpPac5YuB+ud7cTs9iCqjy+8OpM7fIP+rC5vp+VzvJw0M4TfiTX49

RNG Seed: 2428397891

runtime error: invalid memory address or nil pointer dereference
Stack Trace:
goroutine 19 [running]:
runtime/debug.Stack()
	/opt/hostedtoolcache/go/1.18.6/x64/src/runtime/debug/stack.go:24 +0x6
github.com/wowsims/wotlk/sim/core.runSim.func1()
	/home/runner/work/wotlk/wotlk/sim/core/sim.go:55 +0x10
panic({0xe5720, 0x1216e80})
	/opt/hostedtoolcache/go/1.18.6/x64/src/runtime/panic.go:838 +0x29
github.com/wowsims/wotlk/sim/hunter.(*Hunter).registerReadinessCD.func1(0x51442a0, 0x510ea00, 0x5137e00)
[...]

Issue #1261, opened 2022-10-11
Crash Report -1474921766


Ruca777:
Link:
https://wowsims.github.io/wotlk/hunter/#eJztUs9rE0EU3vdmEycjkdGDHXMxXZQGobjZZcF42W29FLwU9A+wYIuHHkJBsMFDEQ/Wk+ilVjBQaxFqVYRqowQiLdRIa0qU1LTag7QhasCe6kn07e7Jm6De/Jhv3uz78c2+4YkER4kmHoNePANZvAL4CHAFoAnQHeeQxBR4cAHuAezTEp8ilyDaO9g33D/klyUKIKR8uoSqsoFG9Mk85ssodNkoouDy+x1Uixv+6cFbVGszjHJvzaIqr1NucQ5rP1Dslo8nUC2QBzSxR96m8OoMM/TXc6iJuMxPohq9zwzm63J5dwpVYxpJ5yFVNSkxSgGSojtnp1DE5PVlNFgrcLSK/v6h7hfW66jm11FodF6k0soLNDhHE6ht/cg11P7jF+Q4B4kKs3B8WJzogRRy2BtEvLzXnRTII8RdxBiREwWREZEI6fGbPgpuxg7wxqXZMWliDib20w40UqyHjQGUgL0EeAfQHt677WaWl3y03BRchcNOp2k6tu1YtknLTNuWbadtx7TTlpPudEwf1kdQfHIL5UQFVWETk6tVTL2qolmrYhP0bRDjcFREhc4b0yywO2MgIoLxrbbg8/Mo+lbeeBZY/vUbWwPN+IKx4J9GBrywde2QdyBsrOxmQs+K2z4QoOZazzG7czEXP9k/2HH63Plc39mk5SRPKf7Px+uy9zdURsoLf6JTsjqGun43OXi9Utt7ryt8PfR+ApLcoCk=

RNG Seed: 1889176454

runtime error: invalid memory address or nil pointer dereference
Stack Trace:
goroutine 11 [running]:
runtime/debug.Stack()
	/opt/hostedtoolcache/go/1.18.6/x64/src/runtime/debug/stack.go:24 +0x6
github.com/wowsims/wotlk/sim/core.runSim.func1()
	/home/runner/work/wotlk/wotlk/sim/core/sim.go:55 +0x10
panic({0xe6640, 0x1231060})
	/opt/hostedtoolcache/go/1.18.6/x64/src/runtime/panic.go:838 +0x29
[...]

PR #1337, opened 2022-10-17
Fix 2 hunter panics

jimmyt857:


jimmyt857:
Fixes #1194 and #1261

PR #1362, opened 2022-10-19
statweight.go changes

JooperGH:
- Changed from a static stat mod of 50 to a relative stat mod of max(50, stat*0.05).
- Added expertise cap detection, both dodge / parry caps are now taken into account.


TheSorm:
I would say with the 5%, better take the maximum of 50 and the 5%. Otherwise you would get really bad results when you have low hit or expertise values for example

PR #1396, opened 2022-10-21
Melee auto attack changes

vigo2:
[core] melee auto attacks have SpellFlagNoOnCastComplete
[warrior] improve t7-4pc handling to only reduce costs for abilities

Quite a few "on successful spellcast" procs aren't checking proc masks at all, and simply assume "OnCastComplete" to imply a (magic) spell cast. This only removes auto attacks, but since setting SpellFlagNoCastComplete is also a very slight perf gain, this seems like a win.

### DIFF
--- a/sim/core/attack.go
+++ b/sim/core/attack.go
@@ -239,7 +239,7 @@ func (unit *Unit) EnableAutoAttacks(agent Agent, options AutoAttackOptions) {
 		ActionID:    ActionID{OtherID: proto.OtherAction_OtherActionAttack, Tag: 1},
 		SpellSchool: unit.AutoAttacks.MH.GetSpellSchool(),
 		ProcMask:    ProcMaskMeleeMHAuto,
-		Flags:       SpellFlagMeleeMetrics | SpellFlagIncludeTargetBonusDamage,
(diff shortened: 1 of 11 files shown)


sim/core/attack.go
```diff
@@ -257,7 +257,7 @@ func (unit *Unit) EnableAutoAttacks(agent Agent, options AutoAttackOptions) {
 		ActionID:    ActionID{OtherID: proto.OtherAction_OtherActionAttack, Tag: 2},
 		SpellSchool: unit.AutoAttacks.OH.GetSpellSchool(),
 		ProcMask:    ProcMaskMeleeOHAuto,
-		Flags:       SpellFlagMeleeMetrics | SpellFlagIncludeTargetBonusDamage,
+		Flags:       SpellFlagMeleeMetrics | SpellFlagIncludeTargetBonusDamage | SpellFlagNoOnCastComplete,
```
jimmyt857, line 260: Should this be applied to ranged autos too?

jimmyt857 (approved):

sim/core/attack.go
vigo2: Doesn't work for ranged autos, since they use AfterCast() to fire OnAutoAttack().

PR #1406, opened 2022-10-22
Refactor dk dots

jimmyt857:


jimmyt857:
Went through each dot 1 by 1 and confirmed each DPS change. Changes in behavior that affected tests:

 - [BUGFIX] Death and decay was not snapshotting the attacker multiplier like every other dot does.
 - ]UNSURE] Wandering plague helper is now applied after the triggering tick is dealt and all aura callbacks are invoked, instead of before other aura callbacks. I.e. this is an ordering change, its now last instead of first.
 - [RNG] No longer using custom outcome function for death and decay, so RNG changed.

Using the new dot logic also fixes the timeline visualization issues.

PR #1410, opened 2022-10-22
Result cache

vigo2:
[core] add a result cache to Spell, speeding up the spell-based damage pipeline somewhat (since each cast would create a new SpellEffect on the heap)

[core] the spellEffect-based damage pipeline now correctly resets Outcome to OutcomeEmpty, so blocks are handled correctly 

[various] related changes (the spell-based damage pipeline now passes *SpellEffect rather than SpellEffect around)

For e.g. DK's dps tests, this reduces heap allocations by >80%, and wowsimwotlk runs >15% faster, consequently.

### DIFF
--- a/sim/core/spell.go
+++ b/sim/core/spell.go
@@ -146,6 +146,8 @@ type Spell struct {
 	initialCritMultiplier           float64
 	initialThreatMultiplier         float64
 	// Note that bonus expertise and armor pen are static, so we don't bother resetting them.
+
+	resultCache SpellEffect
 }
 
 func (unit *Unit) OnSpellRegistered(handler SpellRegisteredHandler) {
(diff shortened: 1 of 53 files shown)


sim/core/spell.go
```diff
@@ -146,6 +146,8 @@ type Spell struct {
 	initialCritMultiplier           float64
 	initialThreatMultiplier         float64
 	// Note that bonus expertise and armor pen are static, so we don't bother resetting them.
+
+	resultCache SpellEffect
```
jimmyt857, line 150: This won't work in the general case, because there are a small number of spells for which there are multiple 'calcs' active at the same time. The examples that come to mind are 1) whenever travel time is long enough and 2) when calculating all aoe spell results before dealing them.

jimmyt857 (changes requested):

sim/core/spell.go
vigo2: I've added an "inUse" bool for SpellEffect that's set in calcDamageInternal(), and reset in dealDamageInternal(). If the resultCache is in use (which only seems to happen for warlocks, with long travel times), a new SpellEffect is created.
I ignored warlock test changes before, because I _always_ get test changes for them on my current setup :/

jimmyt857 (approved):

PR #1439, opened 2022-10-24
Cost reductions

vigo2:
[core] validate cast costs at creation rather than dynamically (in wrapCastFuncResources())

[rogue] static spell cost reduction from 4t7 are now also statically applied (into Spell.BaseCost and Spell.DefaultCast.Cost), so that actual costs are available for rotation planning 

[rogue] the 4t7 cost reduction is now (most likely) modelled correctly, based on additional testing 

[rogue] Feint can now miss, be dodged, be parried or be blocked 

[rogue] the 2t9 dynamic cost reduction is now handled via PseudoStats.CostReduction, and has mostly reliable energy metric 

[warrior] dito for the warrior's 4t7 dynamic cost reduction, which now handles 4t7 procs during Slam casts correctly


b-r-a-n (approved):
Looks good! Should improve the rogue rotation code a bit and the test results seem to indicate that is the case.

PR #1446, opened 2022-10-25
Refactor last 2 paladin spells

jimmyt857:


jimmyt857:
Test changes are just from RNG; crit roll now occurs always, instead of only for non-demon/undead targets.

Issue #1212, opened 2022-10-05
Shadow Priest Rotation


Divinusproelia:
**Describe the bug**
When using the 'Basic' Rotation Type for Shadow Priest, it removed Mind Flay entirely from the logs, rather than removing the 'Clipping' feature as described.

**To Reproduce**
1. Go to Shadow Priest Sim
2. Click on 'Settings'
3. Select 'Basic' under the Rotation Type dropdown
4. Click 'Simulate'
5. Click on 'Detailed Results'
6. Mind flay is no longer in the list of spells

**Expected behavior**
Mind Flay still casts, but all casts are 3 ticks of damage instead of 'clipping' them early to cast other spells.

**Sim Links and Screenshots**
https://wowsims.github.io/wotlk/shadow_priest/#eJztUjtIHFEUnXNnV96OKE+xGK3WRUEE4c0MA1rtaGVhIdpYaqFYWIidVouIRdLYxipdRBBFUHTXH5Ig6yLqgorrmiIgCopMgkXK3LeTJp2FdjnF5Zxzf3Dfs1oEJCnqRC/GMYkM8Bkv5hKwBhwAZ8AD0FMnYCPJVQGGuIrT9UbTRWwWVf0TI9OjUyIh4027sKT8tUP2yXdKVX07pNIWWTG5mSOrRn5k/5h9c29Ly7srskvrZsrUNVIuZ8nO33AXy5CdWjnPTpGdGNcbWnP/Wpl1qHVC/tz5O0vIkFP3GyazZ2an15qd5sgul3V2YTtiMRnmdNzParfA+aMbrc+zOpazlBICCv3xoXj7Ahn/8Q9m+DiS/8AkBqRFQj9/xUmiDR/QqnzlucpzOjo811eup3zH5egojo6vmec7P2CLLwWSqwWyt+8o+ViktqciqeciPcAMUb0IowQjdUuJys7MWFAXbW8IGhc/aeTTXZFzlm4eq+Ay7a5QctAW7/5oc8FbTMnkv77JnNcgulXfbdAd6d/pXgzjDwkJfZc=

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 105.0.5195.127

kevinor:
Just hit this issue; I'm guessing it's pretty low priority since most Shadow Priests going to the trouble of running simulations are probably already clipping to some extent.

SaffatHasan:
[...]

PR #1470, opened 2022-10-26
priest: add known mind flay issue

SaffatHasan:
As described by #1212, add an incorrect simulation of basic rotation to the known issues for Shadow Priest.


SaffatHasan:
As an aside, I left a reference to the issue number, but I did not
provide a link.

This may not be accessible for less technical users, but perhaps
that is inconsequential.

dbyena:
The basic rotation is just bad gameplay. I would prefer removing it from the code entirely. If you review the "ideal" game play/rotation you will see that it's really rather basic.

CyberQuackGames:
Closing this out as Basic rotation has been removed.

PR #1480, opened 2022-10-26
Simplified resists

vigo2:
[core] partial resistance thresholds are no longer cached, which simplifies things in quite a few places (notably, DefenseTables can be removed, and processDynamicBonus() is slimmer)

[core] ResistanceMultiplier() now panics when encountering spells with SpellSchoolNone 

[core] binary(-like) spells (Thorns and Retribution Aura) now have a chance to be fully resisted 

[druid, paladin, shaman] Starfall Splash, Ashtongue Talisman of Zeal, Lightning Bolt and Lavaburst DOTs were accidentally bypassing partial resist checks; they all have a proper spell school set now 

[mage] Frostfire Bolt wouldn't fetch it's cached resistance thresholds, and also bypass resist checks; it's now modelled as encountering 0-resistance mobs (level-based partials only) 

[warrior] simplified Bloodsurge handling by using a SpellFlag on HS, BT, and WW

Performance is unchanged for casters:

```cached

affliction warlock
running 100000 iterations took 28.176578862s
running 100000 iterations took 27.687854948s
running 100000 iterations took 28.28759034s
arcane mage
running 100000 iterations took 10.332333018s
running 100000 iterations took 10.368772998s
running 100000 iterations took 10.439234584s

non-cached

affliction warlock
running 100000 iterations took 27.990349805s
running 100000 iterations took 28.209191846s
running 100000 iterations took 27.950704954s
arcane mage
running 100000 iterations took 10.262150918s
running 100000 iterations took 10.255122587s
running 100000 iterations took 10.376444724s
```


sim/core/spell_resistances.go
```diff
@@ -32,16 +32,24 @@ func (spell *Spell) ResistanceMultiplier(sim *Simulation, isPeriodic bool, attac
 		return attackTable.GetArmorDamageModifier(spell)
 	}
 
+	if spell.SpellSchool.Matches(SpellSchoolNone) {
```
jimmyt857, line 35: Seems like we could put this panic in `unit.RegisterSpell` instead.

jimmyt857 (approved):

sim/core/spell_resistances.go
vigo2: I wasn't sure about healing or other spells, that might just not need a spell school. If you get here, you're a damaging spell.

vigo2: Moved, and guarded the check by DamageMultiplier != 0 || ThreatMultiplier != 0, as for ProcMask. In testing, the only remaining offender was Gift of Arthas, let's hope it stays like this.

jimmyt857: Healing spells actually use both of those multipliers but this is fine to submit. The healing stuff is very WIP still anyway.

PR #1487, opened 2022-10-27
Basic linting preparations

vigo2:
[various] addressed basic linting issues (govet, gofmt, goimport, gosimple, and more)
 - all proto objects are now pointerized (so mutexes aren't copied)
 - various automatic fixes courtesy of gofmt
 - various redundancies removed
 - found some linter issues that are probable bugs (look for "TODO looks fishy")

[deathknight] refactored CastCostPossible()


sim/core/agent.go
```diff
@@ -125,7 +125,7 @@ func (actionID ActionID) ToProto() *proto.ActionID {
 	return protoID
 }
 
-func ProtoToActionID(protoID proto.ActionID) ActionID {
+func ProtoToActionID(protoID *proto.ActionID) ActionID {
```
jimmyt857, line 128: Why are we pointerizing everything? This was done to force the caller to handle the null case. If we're going to make these changes, the null handling logic needs to move with it.

sim/core/statweight.go
```diff
@@ -226,24 +220,26 @@ func CalcStatWeight(swr proto.StatWeightsRequest, statsToWeigh []stats.Stat, ref
 			statMod = meleeHitStatMod
 			if baseStats[stat] < melee2HHitCap && baseStats[stat]+statMod > melee2HHitCap {
 				panic("no")
```
jimmyt857, line 222: I left this debug panic in accidentally from changes yesterday... could you just take it out while you're here  xD

If this is the reason you commented the code, then uncomment it.

vigo2: You already changed it back, i'll just merge.

sim/core/agent.go
vigo2: Pointerizing is because protobuf messages contain a mutex, and mustn't be copied: `copylocks: assignment copies lock value to X: github.com/wowsims/wotlk/sim/core/proto.XXX contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex (govet)`.

I've added an explicit nil-check here now, so it's obvious the caller must check this. Accessing `protoID.Tag` does cause a nil pointer dereference though, anyhow.

jimmyt857 (approved):

PR #1488, opened 2022-10-27
Remove prereqLocation from coldSnap config

### LINKED ISSUE #1442: Mage Talents broken?
Not sure if this is specifically a bug or just something missing, but I was playing with the mage sim and I was readjusting talents to try and get Cold Snap to see what kind of dps increase/loss there might be next to the preset on the sim. Unfortunately despite there being no prerequisites for Cold Snap besides just having 20 points in the frost tree, I cannot get Cold Snap to fill in.

Other talents in the tree light up as available to put a talent point into, but Cold Snap does not. 

image

beltsmith:
Fixes #1442

Cold Snap has no pre-requisites to talent, the existing config set a prereq of coldSnap itself.

### DIFF
--- a/ui/core/talents/mage.ts
+++ b/ui/core/talents/mage.ts
@@ -958,11 +958,7 @@ export const mageTalentsConfig: TalentsConfig<MageTalents> = newTalentsConfig([
 				"spellIds": [
 					11958
 				],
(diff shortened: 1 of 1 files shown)


jimmyt857:
Thanks for the fix!

PR #1492, opened 2022-10-28
Work around expertise rating floating point issues

### LINKED ISSUE #1394: [Warrior] Strength of Arms talent only gives 3 expertise instead of 4
**Describe the bug**
When Strength of Arms talent (in the Arms tree) is talented 2/2, the sim only gives 3 expertise instead of 4.

**To Reproduce**
1. Pick the Strength of Arms talent in the Arms tree
2. Place 2/2 talent points
3. Mouseover the Expertise row in the stats section
4. See that expertise from talents is only 3 (33 rating). 

**Expected behavior**
The talent tooltip shows that it should give 4 expertise when selected 2/2.

**Sim Links and Screenshots**
image
image

beltsmith:
Fixes #1394

Somewhere there is a float precision error happening with expertise rating causing an expertise bonus of 4 to be calculated as 32.78999999 rather than 32.79 (4 * 8.1975). Fixing at two decimals will be rounded up to 32.79 giving the correct expertise of 4

image
image


jimmyt857:
Thanks for the fix! You should hit us up in our discord.

PR #1501, opened 2022-10-28
Balance fixes

Gashiraa:


vigo2:
Changes look good to me. 

Since testing done in https://github.com/JamminL/wotlk-classic-bugs/issues/66 implies that Starfire has a higher proc chance than Wrath, and Aggrend's comment doesn't disagree ("Its core data is to proc 3.5 times per minute"): are you planning to update that part, too?
Also: Hurricane is affected by spell haste, so this could be modelled, too.

vigo2 (changes requested):

vigo2 (approved):

vigo2 (approved):
After gathering more test data, I'm pretty sure Hurricane's chance to proc OOC is slightly higher (I had 705 procs in 45985 casts, or 1.53%), so there's more test results disagreeing with Aggrend's developer quote.
But that's all rather academical ;)

Oh, and if possible: please don't force push, I didn't even notice you had updated the PR.

PR #1505, opened 2022-10-28
Paladin fix seal crit and damage multipliers

Watcher7:


sim/paladin/sov.go
```diff
@@ -200,18 +199,18 @@ func (paladin *Paladin) registerSealOfVengeanceSpellAndAura() {
 }
 
 func (paladin *Paladin) createSealOfVengeanceDot(target *core.Unit) *core.Dot {
+	// DoT is unaffected by 2h spec
+	baseMultiplier := 1 *
```
jimmyt857, line 203: I think this is less readable; why not keep this computation in-line when assigning `DamageMultiplier`? It makes the reader have to go looking for the computation.

Not going to bug you with separate comments, but this applies to pretty much all the cases. Using only `DamageMultiplier` by merging the calc from `DamageMultiplierAdditive` is fine, but none of them are complex enough to warrant separating the calculation from its location of use.

jimmyt857: To give an example of an alterative: you could do something like 
```
DamageMultiplier: 1 *
   (1 +
         x +
         y +
         z) *
   (1 +
         a +
         b +
         c),
```

Watcher7: My thought process was that since some spells proc'd by the same aura share modifiers, then grouping them outside of the specific proc made sense. The DoT's modifiers just got caught up in a pattern. If you want me to go ahead and just move all modifiers into their respective spells I'll update the PR to reflect that.

PR #1514, opened 2022-10-29
Warrior: change spec detection from specific talent to main talent tree

TheGroxEmpire:


sim/warrior/warrior.go
```diff
@@ -29,6 +29,9 @@ type Warrior struct {
 
 	WarriorInputs
 
+	// Main spec talent tree
+	MainSpec string
```
jimmyt857, line 33: There's no need to add an extra field; also string comparison is slower than integer comparison. You can do it like this instead.

```
const (
   ArmsTree = 0
   FuryTree = 1
   ProtTree = 2
)

...

if warrior.PrimaryTalentTree == ArmsTree {
   ...
}
```

TheGroxEmpire: Changed in [1751404](https://github.com/wowsims/wotlk/pull/1514/commits/175140453914b16d2559592c33e134267a3bf37e).

Issue #1552, opened 2022-10-31
Crash Report 518796492


zhoujiaman0:
Link:
https://wowsims.github.io/wotlk/raid/#eJztlM9rE0EUx2feTupm2siYVlxzirlYAoHZXXLwlLR46EEhIEKODbVVQYoUBOkpBi/2YlEqWC/+AeZSPFRXE6iWELWoORhMbRqrQk5aLULxom82P2qpB4XqxXx3mZ3v970ZmN0Py4u9fKqXV1mG+gZT51LjI6PBoxMXzp7SuwQN5CgX4voDMJaqEOpaXIDbReBe8fUNhLTKPHCfuIHFJ1jUVMUnPpTBmPmiNapCzGC1uIxLsZrDZK+oY0MJE4aWoL+GHeWs1vQ+8Rn9lTu4gWrXxTrazKKaFXBWz2qciecO4Dh9T435+2pcd1THUweMwrLyaTe95ECI61TSBEt6LtLwNJCOtmmSWPu5T6cGPUnI+YEkPUP9JACcTNHDUWlbUtqmGbVt00Zj2raMordsMxKxZBSL0rTWqKGnV0GU82C8zkPwWxX6H5VAFkpQp2yd7pmlpEIJr3UA+w+FgB3YCRjjoNPdR0z9w7oHJkZS46PB46nTozoTWmBBAVZztgBTn30bYO+cJmAb8y3AKnNtwNbKTcBw3a8BW3V+BsyrAGvAJcRbx4XLXbzh4vaxDdnVHLRRezjXQk0Xz9qQvXQhy9yFkI6QdSfYcAexnZokg358bSzeM9SVAEWXDsbIFA1btpQyikiZiBFypGjDuRVFxKyItCPSwhibTAXYLQTs0xIY35cg+L4E/TUEbLPaAayjfwgY8RPew0nr+hPnJ2ESSPI+7qc7pFLGGOCNwsGjHirF4zF1RsaIhzGPRkgz3ZLLQ2gFvK5Lj8X3NfK++MHZm0rF2JFG8iJ2aMzVq5iVheAJQ//rNF2O78Yu6eLjXdnnd9R4V8dW4gMNvxkbosP0B5JYn6A=

RNG Seed: 1523560080

Aura Insect Swarm already registered!
Stack Trace:
goroutine 15 [running]:
runtime/debug.Stack()
[...]

Issue #1444, opened 2022-10-25
No option to cast Faerie Fire myself - Boomkin


xaimaway:
It seems to assume 100% uptime of faerie fire cast by another boomkin, I'd like to take into account me being the one casting it please.

lologarithm:
Now if you have adaptive rotation selected and no other armor/hit% debuff selected the balance druid will cast it automatically when down

PR #1556, opened 2022-10-31
Fix IS for raid and have adaptive keep faerie fire up

lologarithm:
Currently multiple moonkin will crash raid sim.

Now with adaptive moonkin will keep up FF if no one else is (in single sim that means if not selected as a debuff)


lologarithm:
fixes https://github.com/wowsims/wotlk/issues/1552 and https://github.com/wowsims/wotlk/issues/1444

PR #1558, opened 2022-11-01
Simpler Major Cooldowns

vigo2:
[core] Unit.HardCast is now also tracking channeled spells (only Expire is set); trying to cast while any spell is casting or channeling fails now
[core] the MajorCooldownManager now honors this one-spell-at-a-time limit
[core] the MCDM doesn't stop early if a major cooldown invoking the GCD is activated

These are all bug fixes, more or less.

[core] MCDM's sortOne() is gone, as a simplification; the perf impact is within the noise
[core] add an early exit for TryUseCooldowns(), using the "minReady" time of MCDs (similar to "minExpires" for the aura tracker)

Since >99% of all calls aren't activating any CDs, this is actually a slight perf win.

[core] simplified MCDM's interface, and...
[diverse] ... changed all clients

Most clients simply wanted (temporary) control over MCDs, and had to do so quite inefficiently.

[hunter] fixed Silencing Shot being fired in the middle of another cast; it should probably become a major cooldown
[rogue] slightly simplified initial disabling and re-enabling of MCDs

### DIFF
--- a/sim/core/cast.go
+++ b/sim/core/cast.go
@@ -23,10 +23,6 @@ type Hardcast struct {
 	Target     *Unit
 }
 
-func (hc *Hardcast) OnExpire(sim *Simulation) {
-	hc.OnComplete(sim, hc.Target)
-}
-
 // Input for constructing the CastSpell function for a spell.
 type CastConfig struct {
 	// Default cast values with all static effects applied.
@@ -65,6 +61,10 @@ type Cast struct {
 	AfterCastDelay time.Duration
 }
 
+func (c *Cast) empty() bool {
+	return c.Cost == 0 && c.GCD == 0 && c.CastTime == 0 && c.ChannelTime == 0 && c.AfterCastDelay == 0
+}
+
 type CastFunc func(*Simulation, *Unit)
 type CastSuccessFunc func(*Simulation, *Unit) bool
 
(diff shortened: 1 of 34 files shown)


sim/core/cast.go
```diff
@@ -65,6 +61,10 @@ type Cast struct {
 	AfterCastDelay time.Duration
 }
 
+func (c *Cast) empty() bool {
+	return c.Cost == 0 && c.GCD == 0 && c.CastTime == 0 && c.ChannelTime == 0 && c.AfterCastDelay == 0
```
jimmyt857, line 65: You can just do `*c == Cast{}`

sim/core/cast.go
```diff
@@ -196,16 +195,29 @@ func (spell *Spell) wrapCastFuncHaste(config CastConfig, onCastComplete CastFunc
 }
 
 func (spell *Spell) wrapCastFuncGCD(config CastConfig, onCastComplete CastFunc) CastFunc {
-	if config.DefaultCast.GCD == 0 {
+	if config.DefaultCast.empty() { // spells that are not actually cast (e.g. auto attacks, procs)
 		return onCastComplete
 	}
 
+	if config.DefaultCast.GCD == 0 { // mostly cooldowns (e.g. nature's swiftness, presence of mind)
+		return func(sim *Simulation, target *Unit) {
```
jimmyt857, line 203: This is going to add a check for every spell, including all the empty-cast spells like autoattacks. Is this a perf loss?

sim/core/cast.go
vigo2: Nope :), this would require an extra `var e Cast` line. Can do that, of course, safer if `Cast` ever changes. Alternatively, though, an `emptyCast` var or the like could be used. `empty()` is used only twice, but it's probably easier to read as a method.

sim/core/cast.go
vigo2: Those are already handled by `DefaultCast.empty()` above - that's true for all autoattacks and procs, at least (which must not be checked here, since they have no such timing restrictions.

jimmyt857 (approved):

PR #1596, opened 2022-11-04
Fix Focus Magic talent calculation of added spell crit %

### LINKED ISSUE #1441: Focus Magic Percent Uptime bug
**Describe the bug**
In your wotlk-simulator, I can not take advantage of "Focus Magic Uptime" when I am not specced inte Focus Fire which is annoying. Let's say that I am playing a Mage-spec that is not using Focus Fire (like frostfire bolt spec) talent but I get the buff from another Mage. Then I should get +3% crit, with that spell cast on me. Now, when I spec into TTW Fire Mage (it gives me 3% crit if the target that I cast the buff on crits), it instead gives me +6% crit which is wrong (it should be +3%). The only way to get 6% crit through this buff is by using the buff on someone else that crits, AND the person using the buff on myself. Right now, I either get 0% if I don't have the buff (which is not always the case since people can use the spell on me even if I don't have the talent) OR 6% by talenting into it (which is not correct either, as it should be 3%)
92e7ba35d265451edfe650c7f5ffe485

**Desktop (please complete the following information):**
 - Chrome

**Additional context**
92e7ba35d265451edfe650c7f5ffe485

PANDAxBEARS:
Fixes #1441

Talenting into Focus Magic was giving an additional static 3% spell crit, it should only add up to 3% based on the uptime configured by the user. The buff focus magic is currently configured via a raid buff in the misc category.

### DIFF
--- a/sim/mage/TestArcane.results
+++ b/sim/mage/TestArcane.results
@@ -9,7 +9,7 @@ character_stats_results: {
   final_stats: 2731.3455
   final_stats: 109
   final_stats: 295.696
-  final_stats: 1580.94502
+  final_stats: 1443.21502
   final_stats: 609
   final_stats: 0
   final_stats: 755.7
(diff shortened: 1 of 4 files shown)


jimmyt857:
Hey thanks for the PR! If you can update the tests again we can merge this.

lologarithm:
At some point we should probably fix focus magic to work correctly in the raid sim. The current way it works is designed only to work in an individual sim.

PR #1621, opened 2022-11-06
Garrote fixes

vigo2:
[rogue] Garrote is no longer half finisher, half builder, it can't crit anymore, may only miss on application, and has it's various multipliers fixed, including multiplying base damage


b-r-a-n (approved):
Looks good; thanks!

PR #1666, opened 2022-11-10
Mage and druid fixes

vigo2:
[core] changed dot.NumberOfTicks and dot.TickCount from int to int32

[mage] Living Bomb's dot application can now miss, and its splash damage scaling corrected
[mage] multiple Living Bomb applications are no longer supported, they were both buggy and effectively unused 
[mage] Ignite is now procced by Living Bomb's crit ticks; cleaned up & simplified 

[druid] Moonfire ticks can no longer miss
[druid] Nature's Grace is now additionally procced by Starfall and Moonfire

I've removed Living Bomb's split into an "applier" and "dot" spell. This makes the "Detailed Results / Damage" section less readable (since misses, hits, and crits are combined), but helps the "Timeline", since it now displays actual tick bars.

Actual dots should probably count "tickHit" and "tickCrit" or the like, so these can be displayed correctly in the "Damage" section - ideally split into "applier" and "dot", using the same SpellID (and icon), so the individual sims can display both, without introducing dummy spells.
 
For Ignite, I've likewise removed the "pseudo-cast". I've no idea how the timeline decides whether to display ticks or not, but with "tickHit" and "tickCrit" support this should work as well.

### DIFF
--- a/sim/mage/living_bomb.go
+++ b/sim/mage/living_bomb.go
@@ -10,15 +10,11 @@ import (
 )
 
 func (mage *Mage) registerLivingBombSpell() {
-
-	actionID := core.ActionID{SpellID: 55360}
-	actionIDDot := core.ActionID{SpellID: 55359} // I want the dot to be separately trackable for metrics
-	actionIDSpell := core.ActionID{SpellID: 44457}
 	baseCost := .22 * mage.BaseMana
 	bonusCrit := float64(mage.Talents.WorldInFlames+mage.Talents.CriticalMass) * 2 * core.CritRatingPerCritChance
 
 	livingBombExplosionSpell := mage.RegisterSpell(core.SpellConfig{
-		ActionID:    actionID,
+		ActionID:    core.ActionID{SpellID: 55362},
 		SpellSchool: core.SpellSchoolFire,
 		ProcMask:    core.ProcMaskSpellDamage,
 		Flags:       SpellFlagMage | HotStreakSpells,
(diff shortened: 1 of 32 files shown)


sim/mage/living_bomb.go
```diff
@@ -29,85 +25,73 @@ func (mage *Mage) registerLivingBombSpell() {
...
-		BonusCritRating:  bonusCrit,
-		DamageMultiplier: mage.spellDamageMultiplier,
-		CritMultiplier:   mage.SpellCritMultiplier(1, mage.bonusCritDamage),
-		ThreatMultiplier: 1 - 0.1*float64(mage.Talents.BurningSoul),
-	})
+	target := mage.CurrentTarget
 
-	mage.LivingBombDots[target.Index] = core.NewDot(core.Dot{
-		Spell: livingBombDotSpell,
+	onTick := func(sim *core.Simulation, target *core.Unit, dot *core.Dot) {
```
jimmyt857, line 67: I haven't tried this yet, but someone added a generic form of the `Ternary` function to our utils (https://github.com/wowsims/wotlk/blob/master/sim/core/utils.go#L130) which would allow you to inline this.

jimmyt857 (approved):

sim/mage/living_bomb.go
vigo2: I'll try a refactoring run using `Ternary` instead of its non-generic siblings. Probably easier to use, and easier to refactor.

PR #1674, opened 2022-11-11
Render rune states

### LINKED ISSUE #448: Merge deathknight runes

Kruhlmann:
Closes #448 

Add rune consolidation and state rendering to the timeline.

Current progress:

image


Kruhlmann:
This PR is still a draft pending refactoring and a new rune display. Ideally the death runes should take the place of the runes they're replacing; not exist separately.

PR #1682, opened 2022-11-12
Rolling dots

vigo2:
[core] added Dot.ApplyOrReset() for rolling dots like Ignite, Deep Wounds, etc.

[hunter] Piercing Shots now correctly resets the tick timer on proc; used Dot.SnapshotBaseDamage to model the rolling damage 

[mage] modelled Ignite like Piercing Shots, since that has much better "Timeline" support 

[warrior] dito for Deep Wounds, changing its base "average weapon damage" now includes all modifiers (dealt and taken), but Deep Wounds itself is only affected by Mangle*

The Deep Wounds "damage buffers" weren't reset in Reset() before, so DW damage was too high. All three dots had extra ticks when proc and tick happened at the same time (since cancelling the previous dot wrongly adds an extra tick in this case). The "damage buffers" were sometimes underflowing, and even producing negative ticks because of this.
"ApplyOrReset()" works around this, but I hope to change this "more fundamentally" in an upcoming patch.

*) @TheGroxEmpire I've updated [the](https://github.com/magey/wotlk-warrior/issues/26) accordingly

### DIFF
--- a/sim/core/dot.go
+++ b/sim/core/dot.go
@@ -74,6 +74,32 @@ func (dot *Dot) Apply(sim *Simulation) {
 	dot.Aura.Activate(sim)
 }
 
+// ApplyOrReset is used for rolling dots that reset the tick timer on reapplication.
+// This is more efficient than Apply(), and works around tickAction.CleanUp() wrongly generating
+// an extra ticks if (re-)application and tick happen at the same time.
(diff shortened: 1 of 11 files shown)


TheGroxEmpire (approved):
Looks to be working as intended.

PR #1689, opened 2022-11-13
More rolling dots, some shaman updates

vigo2:
[deathknight] refactored UnholyBlight similar to e.g. Ignite and Piercing Shots

[paladin] dito for Righteous Vengeance

[shaman] dito for "Electrified" (elemental-4t8)
[shaman] fixed (most) damage multipliers to be additive (all "Apply Aura: Modifiers (Periodic) Damage/Healing" stack like this) 
[shaman] fixed Lava Lash's damage multiplier with FT imbue + enhance-2t8 [shaman] fixed Elemental Precision's threat reduction

### DIFF
--- a/sim/shaman/fire_totems.go
+++ b/sim/shaman/fire_totems.go
@@ -30,7 +30,7 @@ func (shaman *Shaman) registerSearingTotemSpell() {
 			},
 		},
 
-		BonusHitRating:   float64(shaman.Talents.ElementalPrecision) * 1 * core.SpellHitRatingPerHitChance,
+		BonusHitRating:   float64(shaman.Talents.ElementalPrecision) * core.SpellHitRatingPerHitChance,
 		DamageMultiplier: 1 + float64(shaman.Talents.CallOfFlame)*0.05,
 		CritMultiplier:   shaman.ElementalCritMultiplier(0),
 
@@ -106,7 +106,7 @@ func (shaman *Shaman) registerMagmaTotemSpell() {
 			},
 		},
 
-		BonusHitRating:   float64(shaman.Talents.ElementalPrecision) * 1 * core.SpellHitRatingPerHitChance,
+		BonusHitRating:   float64(shaman.Talents.ElementalPrecision) * core.SpellHitRatingPerHitChance,
 		DamageMultiplier: 1 + float64(shaman.Talents.CallOfFlame)*0.05,
 		CritMultiplier:   shaman.ElementalCritMultiplier(0),
 
--- a/sim/shaman/lavaburst.go
+++ b/sim/shaman/lavaburst.go
@@ -19,31 +19,33 @@ func (shaman *Shaman) registerLavaBurstSpell() {
 		0.05*float64(shaman.Talents.Shamanism) +
 		core.TernaryFloat64(shaman.HasMajorGlyph(proto.ShamanMajorGlyph_GlyphOfLava), 0.1, 0)
 
-	applyDot := shaman.HasSetBonus(ItemSetThrallsRegalia, 4)
-	lvbdotDmg := 0.0 // dynamically changing dmg
 	var lvbDot *core.Dot
-	if applyDot {
+	if shaman.HasSetBonus(ItemSetThrallsRegalia, 4) {
 		dotSpell := shaman.RegisterSpell(core.SpellConfig{
 			ActionID:    core.ActionID{SpellID: 71824},
 			SpellSchool: core.SpellSchoolFire,
 			ProcMask:    core.ProcMaskEmpty,
-			Flags:       core.SpellFlagIgnoreModifiers,
+			Flags:       core.SpellFlagNoOnCastComplete | core.SpellFlagIgnoreModifiers,
 
 			DamageMultiplier: 1,
 			ThreatMultiplier: 1,
+
+			ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+				lvbDot.Apply(sim)
+				spell.CalcAndDealOutcome(sim, target, spell.OutcomeAlwaysHit)
+			},
 		})
 		lvbDot = core.NewDot(core.Dot{
 			Spell: dotSpell,
 			Aura: shaman.CurrentTarget.RegisterAura(core.Aura{
 				Label:    "LavaBursted-" + strconv.Itoa(int(shaman.Index)),
-				ActionID: core.ActionID{SpellID: 71824},
+				ActionID: dotSpell.ActionID,
 			}),
 			TickLength:    time.Second * 2,
 			NumberOfTicks: 3,
-			OnSnapshot: func(sim *core.Simulation, target *core.Unit, dot *core.Dot, _ bool) {
-				dot.SnapshotBaseDamage = lvbdotDmg / 3 // spread dot over 3 ticks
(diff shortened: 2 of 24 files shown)


sim/shaman/shocks.go
```diff
@@ -119,16 +118,16 @@ func (shaman *Shaman) registerFlameShockSpell(shockTimer *core.Timer) {
...
 	shaman.FlameShock = shaman.RegisterSpell(config)
 }
 
 func (shaman *Shaman) registerFrostShockSpell(shockTimer *core.Timer) {
 	config := shaman.newShockSpellConfig(49236, core.SpellSchoolFrost, baseMana*0.18, shockTimer)
 	config.Cast.CD.Duration -= time.Duration(shaman.Talents.BoomingEchoes) * time.Second
-	config.DamageMultiplier *= 1 + 0.1*float64(shaman.Talents.BoomingEchoes)
-	config.ThreatMultiplier *= 2
+	config.DamageMultiplier += 0.1 * float64(shaman.Talents.BoomingEchoes)
+	config.ThreatMultiplier *= 2 * shaman.spellThreatMultiplier()
```
jimmyt857, line 130: This would apply `shaman.spellThreatMultiplier()` twice. Is that intended?

jimmyt857 (approved):

sim/shaman/lightning_bolt.go
```diff
@@ -38,25 +38,29 @@ func (shaman *Shaman) newLightningBoltSpell(isLightningOverload bool) *core.Spel
...
+			ActionID:         core.ActionID{SpellID: 64930},
+			SpellSchool:      core.SpellSchoolNature,
+			ProcMask:         core.ProcMaskEmpty,
+			Flags:            core.SpellFlagNoOnCastComplete | core.SpellFlagIgnoreModifiers,
+			DamageMultiplier: 1,
+			ThreatMultiplier: 1,
+
+			ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+				lbDot.ApplyOrReset(sim)
+				spell.CalcAndDealOutcome(sim, target, spell.OutcomeAlwaysHit)
```
lologarithm, line 60: I had it so that this was outside the cast so that way you didn't have to call cast unless the dot wasn't ticking... I guess this way at least you see the exact number of times it was actually applied. I just liked how clean the timeline looked haha

sim/shaman/lightning_bolt.go
```diff
@@ -85,25 +88,19 @@ func (shaman *Shaman) newLightningBoltSpell(isLightningOverload bool) *core.Spel
 		baseDamage := dmgBonus + sim.Roll(719, 819) + spellCoeff*spell.SpellPower()
 		result := spell.CalcDamage(sim, target, baseDamage, spell.OutcomeMagicHitAndCrit)
 
-		if result.Landed() {
```
lologarithm, line 88: If you're going to get rid of the landed check this way, you should just have it exit early. If it did not land then you don't need that check below. Generally speaking I try to write early exits wherever possible as it reduces nesting ifs.

sim/shaman/fire_totems.go
```diff
@@ -30,7 +30,7 @@ func (shaman *Shaman) registerSearingTotemSpell() {
 			},
 		},
 
-		BonusHitRating:   float64(shaman.Talents.ElementalPrecision) * 1 * core.SpellHitRatingPerHitChance,
```
lologarithm, line 33: I just wanted to let you know that all the spots that have `x * 1` was not because I'm a crazy person, the idea was to explicitly show the talents multiplier. Sometimes you'll get like 2% per talent point. If it was 1% per talent point, I would leave the explicit *1

I'm not against you removing them because it is less code, but there was a reason for it 🙂

sim/shaman/lightning_bolt.go
vigo2: This is mostly for uniformity, so all rolling DOTs are implemented (and look) the same way

sim/shaman/fire_totems.go
vigo2: Ye, I've thought about that when removing them. I prefer the shorter version, so I went with that.

sim/shaman/lightning_bolt.go
vigo2: Looking at the code, an early out doesn't make sense since DealDamage() has to called no matter what.

sim/shaman/lavaburst.go
```diff
@@ -79,17 +81,16 @@ func (shaman *Shaman) registerLavaBurstSpell() {
...
-		// TODO: does lava flows multiply or add with elemental fury? Only matters if you had <5pts which probably won't happen.
+		DamageMultiplier: 1 + 0.01*float64(shaman.Talents.Concussion) + 0.02*float64(shaman.Talents.CallOfFlame),
 		CritMultiplier:   shaman.ElementalCritMultiplier([]float64{0, 0.06, 0.12, 0.24}[shaman.Talents.LavaFlows] + core.TernaryFloat64(shaman.HasSetBonus(ItemSetEarthShatterGarb, 4), 0.1, 0)),
-		ThreatMultiplier: 1 - (0.1/3)*float64(shaman.Talents.ElementalPrecision),
+		ThreatMultiplier: shaman.spellThreatMultiplier(),
 
 		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
 			baseDamage := dmgBonus + sim.Roll(1192, 1518) + spellCoeff*spell.SpellPower()
 			result := spell.CalcDamage(sim, target, baseDamage, spell.OutcomeMagicHitAndCrit)
-			if applyDot && result.Landed() {
```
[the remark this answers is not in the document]
vigo2: (That's why I removed the TODO)

sim/shaman/shocks.go
```diff
@@ -104,7 +104,6 @@ func (shaman *Shaman) registerFlameShockSpell(shockTimer *core.Timer) {
 				shaman.LavaBurst.BonusCritRating -= 100 * core.CritRatingPerCritChance
 			},
 		}),
-		// TODO: is this bonus ticks or bonus time that results in extra ticks?
```
vigo2, line 107: (Base duration is 18s, and is shortended by haste. The set adds 9s, and is almost definitely converted into ticks the same way)

sim/shaman/lavaburst.go
```diff
@@ -79,17 +81,16 @@ func (shaman *Shaman) registerLavaBurstSpell() {
 		},
 
 		BonusHitRating:   float64(shaman.Talents.ElementalPrecision) * core.SpellHitRatingPerHitChance,
-		DamageMultiplier: 1 * (1 + 0.01*float64(shaman.Talents.Concussion)) * (1.0 + 0.02*float64(shaman.Talents.CallOfFlame)),
-		// TODO: does lava flows multiply or add with elemental fury? Only matters if you had <5pts which probably won't happen.
```
vigo2, line 83: (Bonus crit damage is always additive)

sim/shaman/lavaburst.go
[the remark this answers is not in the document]
lologarithm: :facepalm:  you are correct

sim/shaman/lightning_bolt.go
lologarithm: I meant if it didn't land you can exit the function (since no damage is applied on a miss)

lologarithm: its ok either way

lologarithm (approved):

PR #1733, opened 2022-11-17
[Stats] Add min/max weapon damage as a stat

b-r-a-n:


proto/common.proto
```diff
@@ -108,6 +108,13 @@ enum Stat {
 	StatFrostResistance = 31;
 	StatNatureResistance = 32;
 	StatShadowResistance = 33;
+	StatRunicPower = 34;
```
jimmyt857, line 111: The stats array is used in a lot of places, so any increase to its size is multiplied greatly in terms of memory footprint. 2 things here:

 1. It doesn't look like the runic power / rune stats are used for anything. They should be removed.
 2. For weapon damage, I think a better approach would be to create a proto for PseudoStats and add these new stats there (the proto doesn't need to match the internal PseudoStats struct, it can just have these two for now). This proto would be passed back and forth along with regular stats in the api, and in the future be expanded to surface some more info to the UI.

proto/common.proto
```diff
@@ -108,6 +108,13 @@ enum Stat {
...
+	StatBloodRune = 35;
+	StatFrostRune = 36;
+	StatUnholyRune = 37;
+	StatDeathRune = 38;
+	StatWeaponDamageMin = 39;
```
jimmyt857, line 116: Rather than 2 separate stats, I think you probably want `weapon_dps` or weapon_damage`. I can't think of any scenario where you'd want to increase one without the other.

PR #1749, opened 2022-11-18
Fix several bugs with explosives

jimmyt857:


jimmyt857:
- Explosive decoy should put other explosives on 2m CD
 - As a result of the previous point, explosive decoy is used in fewer situations.
 - AOE cap applied to explosives.
 - Self-damage from Thermal Sapper Charge added.

darsenault77:
@jimmyt857 I think in line 1186 it should be `< time.Minute*1` rather than `< time.Minute*2`.

If it's NOT less than 1 minute, you would use a Saronite Bomb since it'll have time to get off cooldown and then you would use the Decoy.

For example, on a 180s fight, you would expect
~6s [174s left]: Sapper (on cooldown until 66s)
~66s [114s left]: Saronite (on cooldown until 126s)
~126s [54s left]: Decoy (on cooldown until 246s)

Current condition gives:
~6s [174s left]: Sapper
~66s [114s left]: Decoy (on cooldown until 186s)

jimmyt857:
Sorry for the delay, I did see this but been very busy. Will be fixed in a few minutes.

PR #1767, opened 2022-11-19
Changes to frost sub unh.

enhancedgore:
- Changed condition to enter frost sub unh rotation from HB + points in unh > points in blood to 3/3 BotN, 2/2 Epidemic, 0/3 Frigid Dreadplate.
- Panic when simming frost sub unholy without ERW up (there is no rotation for this it used to just go into ERW opener)
- Frost Sub Unh opener changed to what I think is good (up for debate)
- Changed rotational logic.
- Changed UA checks (UA check 3 is not working)
- Beginning implementation for end of fight rotational optimisation (can be applied to blood as well) (there are changes in frost sub blood)
- Added oblidelay to frost sub unh


rosenrusinov (approved):
I'm not that familiar with the Frost rotations so I'll take your word that this is better in terms of dps. 

What I want you to do in a future PR is clean this up:
- Your method namings dont follow naming convention and needs to be fixed.
- The old rotation code should be moved to a separate file for archival reasons in case we ever need to look back to it for something.
- The raise dead change to IgnoreHaste does nothing, you should be careful with pushing changes that were just tests for you in your actual PRs

@JooperGH press the merge button if you think this is good to go

PR #1872, opened 2022-11-21
Use maps for remaining database fields

jimmyt857:


jimmyt857:
DB changes are because the enchants are now sorted by EffectID

PR #1890, opened 2022-11-22
Scorch hot streak proc

m-weeks:
Scorch can trigger hot streak as well, but it was not flagged as being able to do so


jimmyt857:
Thanks for the fix!

Issue #1897, opened 2022-11-22
Suggestion: Homepage sim links improvements


kayla-glick:
# Overview

The homepage's various sim links look a bit cluttered/disorganized rght now because of the varying widths and positioning. It could be improved by taking an approach similar to the dropdown from the individual sim pages. In addition, there's now 17 unique spec/class links in addition to the raid sim link. In fact, not all of these are reflected in the individual sim page's dropdown list. I'd like to suggest some improvements to make it look nicer.

### Approach
As mentioned, there's sort of 2 issues I see right now. The first being the "cluttered" looks of the different links because of their varying sizing/positioning, and the second being the sheer number of links now. I'd like to suggest the following changes:

1. Format the class/spec sim links to be equally-sized blocks similar to the dropdown on the individual sim page.
2. Use 10 class blocks instead of the 17 unique class/spec sim blocks. For classes with multiple pages, have them toggle a dropdown where the specific sim can be opened.
  a. Each block would use the class icon
  b. 10 class blocks is a nice round number and could be organized into 2 rows of 5 or 5 rows of 2
  c. Blocks could also be styled to use white text with colorful class color backgrounds similar to the individual sim dropdown (not seen in image below)

**Quick example mock-up**
<img width="1728" alt="image" src="https://user-images.githubusercontent.com/12898988/203346477-5e9f5810-f3eb-40a0-8e04-9e7d590259cb.png">

Issue #1896, opened 2022-11-22
Suggestion: Social Links (GitHub, Discord, Patreon)


kayla-glick:
# Overview
Links to the Discord server, GitHub repo, and Patreon could help increase engagement and support, and are pretty standard parts of open-source projects. Let's advertise them! 😄 

### Homepage

Links could be added below the various class sim links as seen below. It's important to provide a larger margin between the two to provide separation.

<img width="1728" alt="Screen Shot 2022-11-22 at 9 25 54 AM" src="https://user-images.githubusercontent.com/12898988/203339074-2315ec38-1fce-41c9-a3c5-c616401c4143.png">

## Sim UI
This one would be a bit harder to fit into the UI, but I think we can do it. There's a couple of different ways you could approach it.

**Option 1**
There's already a Patreon link in the right-hand top toolbar, so it's a perfect place to include them! We'd want to provide a little bit of separation from the more functional buttons with a small divider.

<img width="1728" alt="Screen Shot 2022-11-22 at 9 34 52 AM" src="https://user-images.githubusercontent.com/12898988/203341509-e125faff-f118-4f7b-ad8a-29e304eed086.png">

**Option 2**
We could add links in a separate footer element that's accessible via scrolling. This would give us the most room to work and is probably most commonly used. It would enable us to include the logo/homepage link and other content as well.

<img width="1728" alt="Screen Shot 2022-11-22 at 9 42 18 AM" src="https://user-images.githubusercontent.com/12898988/203342927-cc6085b9-04e2-4ecf-9452-73b6446c670b.png">

**Option 3**
Socials could be fit in below the left-hand stats table. The problem with this approach is that many specs have significantly larger stat tables that affect them, and already have shrunken font sizes to deal with spacing issues.

<img width="1728" alt="Screen Shot 2022-11-22 at 9 29 45 AM" src="https://user-images.githubusercontent.com/12898988/203340043-12596682-e703-49a2-b607-e484490e18ff.png">

lologarithm:
My vote would be the homepage probably just to keep as much space as possible in the sim for sim stuff. Header bar with patreon link would be second best imo

PR #1912, opened 2022-11-23
UI: Homepage Rework

kayla-glick:
In https://github.com/wowsims/wotlk/issues/1897 and https://github.com/wowsims/wotlk/issues/1896 I proposed some changes to the homepage to improve usability and advertise social links for Discord, GitHub, and Patreon. After playing around with styles for a bit I ultimately settled on this format. The page's contents are now a bit better organized, and the 17 class sim links have been condensed so that they're grouped by class. Hovering or clicking on a class reveals a "dropdown" menu with the links for classes with more than 1.

With this I also started a rework of our shared typescript and sass files, creating `ui/scss` and `ui/shared`. I also added PopperJS and Bootstrap Node Modules and imported them only for the Homepage for now, moving us away from relying on CDNs and giving us more control over customizing content from those libraries. One awesome bonus is that we'll now have thematic Bootstrap classes for all classes! You can now use `*-classname` classes or css variables to stylize content with class colors!

### Homepage - Large Screen
image

Individual spec sims can be revealed by hovering, clicking, or pressing a class link
image

### Mobile optimized (Extra small through medium sizes)
image
image
image
image


kayla-glick:
TODO: Get a permanent Discord invite link

PR #1924, opened 2022-11-25
reorganizing and streamlining stylesheets

kayla-glick:
Heh sorry that this turned out so large 😅 in order to swap 1 spec I had to swap all of them, so I didn't really have a choice with it.

Stylesheets were previously stored all over the place, making them hard to find at times. I tried to streamline them so that they'll live in a single directory `ui/scss`. Ideally `scss` lives in the root directory but changes wouldn't be detected at the moment when running `make $spec`.

In addition, each new sim type has their own directory inside `ui/scss/sims` where styles can be assigned. `ui/scss/sims/<simname>/index.scss` is the default stylesheet loaded for any given sim.

The new directory structure looks like this:

```
> ui
  > scss
    > core
      > components
        ...
      > individual_sim_ui
        ...
      > sim_ui
        ...
      > talents
        ...
    > homepage
      ...
    > sims
      > <simname> // E.g. balance_druid
        > index.scss
        ...
    ... other global scss files
```

This should be a big step in streamlining our stylesheets and should make development of the v3.0 UI.


kayla-glick:
Manually tested each sim type to make sure that they were all working after these changes. `make $spec` often catches import/naming issues as well which was handy

jimmyt857:
The previous structure was to store stylesheets in the same directory as their corresponding .ts file. I'm not seeing the advantage of the new structure, could you clarify?

kayla-glick:
Hey @jimmyt857! This is just the general convention that I've seen used in web apps and online when I've looked into SCSS storage in Node apps. People typically centralize their CSS and SCSS files similar to images and other assets. If there's any benefit I get it would be not worrying about name conflicts with the typescript files 😃

kayla-glick:
Going to add back `_` prefixes to denote Sass partials

Kruhlmann:
I agree with @kayla-glick in regards to the structure. I would also prefer to keep the SCSS files in their own directory, and I have a similar experience with other projects as hers.

jimmyt857:
Ok, SGTM

kayla-glick:
@jimmyt857 @Kruhlmann appreciate the feedback! This should be ready to go now. I added back sass partial naming scheme

PR #1933, opened 2022-11-25
fix: blood tap logic

lime-green:
Fixes 3 things:
1. Blood Tap refreshes the CD on both blood or death runes, whichever rune comes first. Not just deaths.
2. Similarly, Blood Tap converts the first blood or death rune, fixed the code where it was prioritizing active runes
3. Using Blood Strike / Pestilence (Blood of the North) on a blood-tapped Death Rune should not cause it to stay as a Death rune when Blood Tap expires, this has been fixed.

Removes revertOnSpend because the attribute is no longer needed

Before:
Notice how both death runes regenerate back after the Blood Tap
image

After:
Only one death rune regenerates back after the Blood Tap, the other is blood again
image

### DIFF
--- a/sim/core/runic_power_helper.go
+++ b/sim/core/runic_power_helper.go
@@ -143,6 +143,8 @@ func (rp *RunicPowerBar) CorrectBloodTapConversion(sim *Simulation, bloodGainMet
 	if slot > -1 {
 		rp.btslot = slot
 		rp.ConvertToDeath(sim, slot, sim.CurrentTime+time.Second*20)
+		rp.GainRuneMetrics(sim, rp.deathRuneGainMetrics, 1)
+		rp.onDeathRuneGain(sim)
 	}
 
 	slot = -1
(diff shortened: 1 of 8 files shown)


lime-green:
I'm going to play around with this a bit more and see if I can detect any other issues

sim/core/runic_power_helper.go
```diff
@@ -134,7 +134,7 @@ func (rp *RunicPowerBar) CorrectBloodTapConversion(sim *Simulation, bloodGainMet
 	// so in english
 	// 1. try to convert active blood rune -> death rune
 	// 2. if no active blood, convert inactive blood rune -> death rune
-	// 3. then convert one inactive death rune -> active
+	// 3. then convert one inactive blood or death rune -> active
 
 	slot := int8(-1)
 	if rp.runeStates&isSpentDeath[0] == 0 {
```
lime-green, line 140: I want to confirm this logic, I’m pretty sure what happens in game is it takes the first blood and converts it, not sure it prioritized active runes

sim/paladin/protection/rotation.go
```diff
@@ -19,7 +19,7 @@ func (prot *ProtectionPaladin) OnGCDReady(sim *core.Simulation) {
 func (prot *ProtectionPaladin) customRotation(sim *core.Simulation) {
 	// Setup
 	target := prot.CurrentTarget
-	
+
```
lime-green, line 22: Everything here was changed by go fmt

sim/core/runic_power_helper.go
lime-green: confirmed in game and fixed code. Blood Tap just goes from left to right and checks for the first rune off cooldown, it doesn't prioritize active runes

PR #1955, opened 2022-11-27
Update sim title dropdown + part of sidebar

kayla-glick:
### Reference image
image

### Actual screenshots
image
image
image
image


ui/core/components/sim_title_dropdown.ts
```diff
@@ -0,0 +1,262 @@
...
+  type: 'Spec'
+  index: Spec
+}
+
+interface RaidOptions {
+  type: 'Raid'
+}
+
+// Dropdown menu for selecting a player.
+export class SimTitleDropdown extends Component {
```
kayla-glick, line 30: Completely rewrote this component

ui/shared/bootstrap_overrides.ts
```diff
@@ -1,26 +1,40 @@
 import * as bootstrap from 'bootstrap';
```
kayla-glick, line 1: I rewrote these handlers because they weren't being applied on the SimUI page because the title dropdown loads in via JavaScript. This is more inline with how Bootstrap handlers actually work

Issue #1611, opened 2022-11-05
Mana potion injector should give more mana to engineering


lologarithm:
When a user has engineering selected Mana Potion Injector should give more mana. (and health injector should give more health)

PR #1994, opened 2022-12-01
add runic injector potions

lime-green:
Fixes https://github.com/wowsims/wotlk/issues/1611


sim/core/consumes.go
```diff
@@ -514,6 +517,9 @@ func registerPotionCD(agent Agent, consumes *proto.Consumes) {
 				if defaultPotion == proto.Potions_RunicManaPotion {
 					character.ExpectedBonusMana -= float64((4200 + 4400) / 2)
 				}
+				if defaultPotion == proto.Potions_RunicManaInjector {
+					character.ExpectedBonusMana -= (4200 + 4400) / 2 * 1.25
```
lologarithm, line 521: do you need the isEngi conditional on the x1.25 there?

lime-green: yep, good catch, fixed

PR #2011, opened 2022-12-02
Move timeline scrollbar to be visible all the time, without scrolling all the way down

kayla-glick:
The main problem with this however is that the labels are hidden by scrolling.
image


lologarithm:
can we make the labels on the left and the timebar at the top all outside the timeline itself?

PR #2019, opened 2022-12-03
Fix explosive CD logic and allow each explosive type to be separately…

jimmyt857:
… specified in CD timings


jimmyt857:
Also addresses condition bug in #1749

PR #2090, opened 2022-12-08
Frost Sub-Blood Desync Rotation

lime-green:
Improves desync rotation by 200-400dps (tested in p2 gear, with 2-4 minute fight lengths)


sim/core/runic_power.go
```diff
@@ -365,7 +365,10 @@ func (rp *RunicPowerBar) BloodRuneBothReadyAt() time.Duration {
 
 }
 
-func (rp *RunicPowerBar) RuneReadyAt(slot int8) time.Duration {
+func (rp *RunicPowerBar) RuneReadyAt(sim *Simulation, slot int8) time.Duration {
```
lologarithm, line 368: thanks for this change, we were just talking about this on the discord!

rosenrusinov (approved):

PR #2098, opened 2022-12-08
Add Demonic Pact Average SP to the topline of 'Results' tab for demo locks only

ptierney13:


ptierney13:
Closed temporarily to make some updates

PR #2176, opened 2022-12-16
warlock: implement infernal

Tharre:


sim/warlock/inferno.go
```diff
@@ -0,0 +1,203 @@
...
+		TickLength:          time.Second * 2,
+		AffectedByCastSpeed: false,
+		OnTick: func(sim *core.Simulation, target *core.Unit, dot *core.Dot) {
+			// TODO: use highest SP amount of all schools
+			// base formula is 25 + (lvl-50)*0.5 * Warlock_SP*0.2
+			// note this scales with the warlocks SP, NOT with the pets
+
+			// we remove all the spirit based sp since immolation aura doesn't benefit from it, see
+			// JamminL/wotlk-classic-bugs#329
+			coef := core.TernaryFloat64(infernal.owner.GlyphOfLifeTapAura.IsActive(), 0.2, 0) +
```
jimmyt857, line 168: Looks like this coeff can be computed outside the OnTick function for better performance.

PR #2220, opened 2022-12-21
ee: dont overwrite same priority with less duration

jarveson:
Signed-off-by: jarves <jarveson@gmail.com>

### DIFF
--- a/sim/core/exclusive_effect_test.go
+++ b/sim/core/exclusive_effect_test.go
@@ -1,0 +1,56 @@
+package core
+
+import (
+	"testing"
+	"time"
+)
+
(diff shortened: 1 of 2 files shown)


sim/core/exclusive_effect_test.go
```diff
@@ -0,0 +1,56 @@
+package core
+
+import (
+	"testing"
+	"time"
+)
+
+func TestSingeAuraExclusiveDurationNoOverwrite(t *testing.T) {
```
jimmyt857, line 8: Typo, 'Singe' instead of 'Single'. Same for the other test name.

jimmyt857 (approved):

PR #2241, opened 2022-12-29
Weapon Effects support for Weapon Swapping

Horatio27:
Working on modeling Weapon enchants to use the WeaponEffect instead of EnchantEffect.

### DIFF
--- a/sim/common/wotlk/enchant_effects.go
+++ b/sim/common/wotlk/enchant_effects.go
@@ -162,33 +162,55 @@ func init() {
 		character.PseudoStats.ThreatMultiplier *= 0.98
 	})
 
-	core.AddWeaponEffect(3789, func(agent core.Agent, slot proto.ItemSlot) {
+	core.NewEnchantEffect(3789, func(agent core.Agent) {
 		character := agent.GetCharacter()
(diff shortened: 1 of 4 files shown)


sim/common/wotlk/enchant_effects.go
```diff
@@ -162,38 +162,31 @@ func init() {
...
-			Label:    "Berserking (Enchant)",
+			Label:    name + " Berserking (Enchant)",
 			Duration: core.NeverExpires,
 			OnReset: func(aura *core.Aura, sim *core.Simulation) {
 				aura.Activate(sim)
 			},
 			OnSpellHitDealt: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
-				if !result.Landed() || !spell.ProcMask.Matches(core.ProcMaskMelee) {
+				hasEnchant := character.Equip[slot].Enchant.EffectID == 3789
+				if !hasEnchant || !result.Landed() || !spell.ProcMask.Matches(core.ProcMaskMelee) {
```
jimmyt857, line 184: This check works well enough for cases like this, but there's no way here to activate any ICDs, which will be needed for other effects.

The vague idea I had about this is that `AddWeaponEffect` could have a return value `type OnSwapItem(slot ItemSlot)` so that this code can provide a handler for item swaps. Then `core` would need to store that function somewhere and call it at the right time.

PR #2247, opened 2022-12-30
OnCastComplete rework

Tharre:


sim/druid/balance/TestBalance.results
```diff
@@ -46,911 +46,911 @@ character_stats_results: {
...
  key: "TestBalance-AllItems-WrathfulGladiator'sIdolofResolve-51429"
  value: {
-  dps: 7355.53047
-  tps: 7194.44067
+  dps: 7320.79478
+  tps: 7159.70498
  }
 }
 dps_results: {
  key: "TestBalance-Average-Default"
```
jimmyt857, line 902: Looks like balance is losing 34 DPS in the average test. Is this expected?

sim/druid/talents.go
```diff
@@ -314,14 +314,15 @@ func (druid *Druid) applyOmenOfClarity() {
 				spell.CostMultiplier += 1
 			}
 		},
-		OnCastComplete: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell) {
```
jimmyt857, line 317: Instead of deleting this, I think we could have a local variable `var proccedAt time.Duration` which is set on proc, and then only remove the aura if `proccedAt != sim.CurrentTime`.

Then you'll be able to remove the hard-coded handling of clearcasting from all the spells.

sim/paladin/shield_of_righteousness.go
```diff
@@ -9,6 +9,8 @@ import (
 
 func (paladin *Paladin) registerShieldOfRighteousnessSpell() {
 	baseCost := paladin.BaseMana * 0.06
+	has4pT8 := paladin.HasSetBonus(ItemSetAegisPlate, 4)
+	procAura := paladin.NewTemporaryStatsAura("Aegis", core.ActionID{SpellID: 64883}, stats.Stats{stats.BlockValue: 225}, time.Second*6)
```
jimmyt857, line 13: Could we refactor this slightly like so?

```
var aegisPlateProcAura *core.Aura
if paladin.HasSetBonus(ItemSetAegisPlate, 4) {
  aegisPlateProcAura = paladin.NewTemporaryStatsAura("Aegis", core.ActionID{SpellID: 64883}, stats.Stats{stats.BlockValue: 225}, time.Second*6)
}

...
...

if aegisPlateProcAura != nil {
  aegisPlateProcAura.Activate(sim)
}
```

The advantage is that the aura won't be registered unless the set bonus is equipped, so in most cases we'll have 1 less aura attached to the unit.

sim/priest/items.go
```diff
@@ -89,22 +89,22 @@ var ItemSetSanctificationRegalia = core.NewItemSet(core.ItemSet{
 		2: func(agent core.Agent) {
 		},
 		4: func(agent core.Agent) {
-			priest := agent.(PriestAgent).GetPriest()
```
jimmyt857, line 92: I think you can revert this change entirely. It seems likely that activating after the PWS cast is correct behavior.

sim/warlock/lifetap.go
```diff
@@ -46,6 +47,14 @@ func (warlock *Warlock) registerLifeTapSpell() {
 			if hasGlyph {
 				warlock.GlyphOfLifeTapAura.Activate(sim)
 			}
+
+			if has4PT7 {
+				if warlock.SpiritsoftheDamnedAura.IsActive() {
```
jimmyt857, line 52: `aura.Activate()` already checks for `IsActive()` and does the refresh. You can simplify this code here, which will also cause the PPM metrics for the aura to be correct.

sim/warlock/items.go
```diff
@@ -80,15 +80,15 @@ var ItemSetPlagueheartGarb = core.NewItemSet(core.ItemSet{
 				OnReset: func(aura *core.Aura, sim *core.Simulation) {
 					aura.Activate(sim)
 				},
-				OnCastComplete: func(aura *core.Aura, sim *core.Simulation, spell *core.Spell) {
```
jimmyt857, line 83: Seems like this change is a no-op? The new implementation in `life_tap.go` also applies the aura AFTER the calculation.

When reasonable, keeping the set bonus logic isolated to the set bonus function is preferred because its easier for the reader to understand.

sim/druid/balance/TestBalance.results
Tharre: If you look at the Remaining Differences commit, balance damage doesn't change at all there, so it should be fully explained by the caster trinket change.

sim/druid/talents.go
Tharre: I don't think that would work. The issue here why we have to change anything at all is that we have a CC proc -> we cast something for free -> OnSpellHitDealt is called and may proc another CC -> OnCastComplete is called and removes the CC proc. Weirdly even when moving the check into ApplyEffects the DPS is still different, I have no clue yet why that is.

Thinking about it, maybe we can check in OnCastComplete if the aura was applied at the exact same time as sim.CurrentTime and only remove it if they differ.

sim/paladin/shield_of_righteousness.go
Tharre: Yeah that makes more sense.

sim/priest/items.go
Tharre: Agreed.

sim/warlock/lifetap.go
Tharre: Oh, nice.

sim/warlock/items.go
Tharre: It is, yes. The only purpose of that commit was to prove that warlock DPS doesn't change from any other interaction.

PR #2257, opened 2023-01-01
Item swap core

Horatio27:
Implementing the Item Swapping into core.

### DIFF
--- a/sim/core/item_swaps.go
+++ b/sim/core/item_swaps.go
@@ -30,6 +30,10 @@ type ItemSwap struct {
 	has2H bool
 }
 
+/*
+	TODO All the extra parameters here and the code in multiple places for handling the Weapon struct is really messy,
+		we'll need to figure out something cleaner as this will be quite error-prone
+*/
 func (character *Character) EnableItemSwap(itemSwap *proto.ItemSwap, mhCritMultiplier float64, ohCritMultiplier float64, rangedCritMultiplier float64) {
 	items := getItems(itemSwap)
 


sim/common/wotlk/enchant_effects.go
```diff
@@ -197,6 +195,22 @@ func init() {
...
+			oh = character.Equip[proto.ItemSlot_ItemSlotOffHand].Enchant.EffectID == 3789
+
+			procMask = core.GetMeleeProcMaskForHands(mh, oh)
+			ppmm = character.AutoAttacks.NewPPMManager(1.0, procMask)
+			if ppmm.Chance(procMask) == 0 {
+				aura.Deactivate(sim)
+				return
+			}
+
+			if !aura.IsActive() {
```
jimmyt857, line 210: This check is not necessary, `Active` already handles the case where the aura isn't active.

sim/common/wotlk/enchant_effects.go
```diff
@@ -197,6 +195,22 @@ func init() {
...
+
+		core.RegisterOnItemSwap(3789, func(sim *core.Simulation) {
+			mh = character.Equip[proto.ItemSlot_ItemSlotMainHand].Enchant.EffectID == 3789
+			oh = character.Equip[proto.ItemSlot_ItemSlotOffHand].Enchant.EffectID == 3789
+
+			procMask = core.GetMeleeProcMaskForHands(mh, oh)
+			ppmm = character.AutoAttacks.NewPPMManager(1.0, procMask)
+			if ppmm.Chance(procMask) == 0 {
+				aura.Deactivate(sim)
+				return
```
jimmyt857, line 207: This `return` could be removed and replaced with an `else` clause. Same comment for many of the other effects.

Guard clauses (https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) are better for readability when there are several different cases, but when there are only 2 its usually easier to understand with an if-else.

sim/core/database.go
```diff
@@ -188,6 +188,15 @@ func ProtoToEquipmentSpec(es *proto.EquipmentSpec) EquipmentSpec {
 	return coreEquip
 }
 
+func ProtoToItem(itemSpec *proto.ItemSpec) Item {
```
jimmyt857, line 191: This function is not necessary, you can just call `NewItem(*itemSpec)` instead.

sim/core/item_swaps.go
```diff
@@ -0,0 +1,170 @@
...
+	newItem := swap.GetItem(slot)
+
+	// No item to swap too
+	if newItem.ID == 0 {
+		return false
+	}
+
+	character.Equip[slot] = newItem
+	stats := newItem.Stats.Add(oldItem.Stats.Multiply(-1))
+	character.AddStatsDynamic(sim, stats)
```
jimmyt857, line 124: `AddStatsDynamic` is fairly expensive because it triggers a bunch of other things; if possible, it would be best to call it only once by adding the stats for all items together.

sim/core/item_swaps.go
```diff
@@ -0,0 +1,170 @@
...
+
+import (
+	"time"
+
+	"github.com/wowsims/wotlk/sim/core/proto"
+)
+
+type OnSwapItem func(*Simulation)
+
+var onSwapCallbacks = map[int32]OnSwapItem{}
```
jimmyt857, line 11: Move this into the `ItemSwap` struct and make it a list instead of a map (no ID key). Then just call all swap functions registered to the character.

sim/common/wotlk/enchant_effects.go
```diff
@@ -228,6 +242,23 @@ func init() {
 				}
 			},
 		})
+
+		core.RegisterOnItemSwap(3241, func(sim *core.Simulation) {
```
jimmyt857, line 246: Since this exact logic is repeated so much, I think you should be able to add a helper function in `item_swap.go` that reduces this whole block to a 1-line call.

sim/common/wotlk/enchant_effects.go
```diff
@@ -380,39 +379,25 @@ func init() {
...
-						return
-					}
-				} else if mh && oh {
-					if !spell.ProcMask.Matches(core.ProcMaskMelee) {
-						return
-					}
-				}
-
 				vulnAura.Activate(sim)
-				isMH := spell.ProcMask.Matches(core.ProcMaskMeleeMH)
```
jimmyt857, line 414: This looks like a bug, `isMH` and `isOH` have the same value

sim/core/item_swaps.go
```diff
@@ -0,0 +1,247 @@
...
+		character.AutoAttacks.OH = character.WeaponFromOffHand(swap.ohCritMultiplier)
+		//Special case for when the OHAuto Spell was setup with a non weapon and does not have a crit multiplier.
+		character.AutoAttacks.OHAuto.CritMultiplier = swap.ohCritMultiplier
+		break
+	case proto.ItemSlot_ItemSlotRanged:
+		character.AutoAttacks.Ranged = character.WeaponFromRanged(swap.rangedCritMultiplier)
+		break
+	}
+
+	character.AutoAttacks.IsDualWielding = character.Equip[proto.ItemSlot_ItemSlotMainHand].SwingSpeed != 0 && character.Equip[proto.ItemSlot_ItemSlotOffHand].SwingSpeed != 0
```
jimmyt857, line 227: I don't see any shield support, but you'll want to also set `character.PseudoStats.CanBlock` here.

sim/core/item_swaps.go
```diff
@@ -0,0 +1,247 @@
...
+
+		if !mh && !oh {
+			aura.Deactivate(sim)
+		} else {
+			aura.Activate(sim)
+		}
+	})
+}
+
+func (character *Character) EnableItemSwap(itemSwap *proto.ItemSwap, mhCritMultiplier float64, ohCritMultiplier float64, rangedCritMultiplier float64) {
```
jimmyt857, line 83: All the extra parameters here and the code in multiple places for handling the Weapon struct is really messy. It's OK for now if you want to get this submitted soon, but we'll need to figure out something cleaner as this will be quite error-prone.

PR #2388, opened 2023-01-13
Refactor DK Costs and remove RuneSpell

jimmyt857:


sim/deathknight/runespell.go
```diff
@@ -71,18 +70,8 @@ func (rs *RuneSpell) castInternal(sim *core.Simulation, target *core.Unit) bool
...
-
 	rs.DoCost(sim)
 
-	if rs.onCast != nil {
-		rs.onCast(sim)
-	}
-
-	if rs.dk.BladeBarrierAura.IsActive() {
-		rs.dk.BladeBarrierAura.Refresh(sim)
-	}
```
rosenrusinov, line 84: BladeBarrier needs to be refreshed on every DK spell cast while there are 0 blood runes. I'm not sure it had the 0 runes check before but it should be checked on every cast which i guess makes sense to be kept here

rosenrusinov (commented):
1 minor issue with blade barrier

sim/deathknight/talents_blood.go
```diff
@@ -295,12 +295,6 @@ func (dk *Deathknight) applySuddenDoom() {
 
 	procChance := 0.05 * float64(dk.Talents.SuddenDoom)
 
-	dk.SuddenDoomAura = dk.RegisterAura(core.Aura{
-		Label:    "Sudden Doom Proc",
-		ActionID: core.ActionID{SpellID: 49530},
-		Duration: core.NeverExpires,
-	})
```
rosenrusinov, line 302: This is used in the result metrics to know how many times you've procced the aura so it should be reverted even if not used by the GCD/Cost skip mechanics anymore

rosenrusinov:
I've been doing a lot of Blood DK related updates in the past days so you should update with master as you already have conflicts with my work (not that many tho)

sim/deathknight/runespell.go
jimmyt857: Added this back now

sim/deathknight/talents_blood.go
jimmyt857: Put this back too

sim/deathknight/heart_strike.go
```diff
@@ -56,39 +66,15 @@ func (dk *Deathknight) newHeartStrikeSpell(isMainTarget bool, isDrw bool) *RuneS
 			}
 		},
 	}
-	if isDrw {
-		conf.DamageMultiplier *= .5
-		conf.Flags |= core.SpellFlagIgnoreAttackerModifiers
-	}
```
rosenrusinov, line 62: This seems to have gone and we still need to apply those things to both main and off target DRW spells

jimmyt857: Fixed this, nice catch!

PR #2501, opened 2023-01-26
fix leftover experimental settings still being used

lime-green:
Previously if you had an experimental setting and turned off the setting it would still use the setting value, but it shouldn't


ui/core/proto_utils/utils.ts
```diff
@@ -159,6 +159,26 @@ export const naturalClassOrder: Array<Class> = [
 	Class.ClassWarrior,
 ]
 
+export const ExperimentalRotationSettings = new Set([
+	'preNerfedGargoyle',
+	'desyncRotation',
+])
```
kayla-glick, line 165: Really not a fan of hardcoding these here. We should be able to set a given input as experimental when defining them in a sim's `inputs.ts` files

kayla-glick (changes requested):

ui/core/proto_utils/utils.ts
lime-green: I agree, fixed

rosenrusinov:
@kayla-glick can you please review your request for changes

Issue #2567, opened 2023-02-02
Rogue - Mutilate used while not talented


gewoonnico:
**Describe the bug**
Even though I'm using the Hybrid AoE build; when simming on <4 targets and wielding daggers, Mutilate is simmed instead of Sinister Strike.
I'm assuming there is no check on the talent implemented here, but rather a check on weapon type equipped?

**To Reproduce**
1. Open the Rogue sim and setup [talents for FoK](https://www.wowhead.com/wotlk/talent-calc/rogue/005320105352100520103-02520520500310102221_001qf511xrp21xrr) (link is just an example)
2. Equip two daggers as weapons
3. Use less than 4 targets in encounter settings (so FoK is not used)
4. Run a simulation and go to the Results tab
  a. Mutilate is used, even though it shouldn't be possible

**Expected behavior**
Sinister Strike is used when Mutilate is not talented.

**Sim Links and Screenshots**
[...]

PR #2568, opened 2023-02-02
Fix detection for Rogue Mutilate availability

gewoonnico:
Use complete `rogue.CanMutilate()` check method instead of incomplete in-line check.


rosenrusinov (approved):

rosenrusinov:
Fixes #2567

PR #2601, opened 2023-02-05
Encounter AI updates and Hodir

rosenrusinov:
- added support for critical damage multiplier buffs like storm power in hodir encounter
- Made changes to the target ai to support it running in individual sims without tank targets
- Made sure all our encounters work with the changes
- Added a new encounter preset for Hodir
- Implemented most of the hodir encounter abilities and added basic ai for both individual and raid sims (still a work in progress)


dbyena:
When do you expect on getting this merged?

rosenrusinov:
> When do you expect on getting this merged?

Hopefully soon. I need someone to look into the front end part of the PR as im not a front end developer so might be a way to do it better

sim/core/unit.go
```diff
@@ -160,6 +160,27 @@ func (character *Character) DoNothing() {
 	character.doNothing = true
 }
 
+func (unit *Unit) IsActive() bool {
```
jimmyt857, line 163: `IsEnabled` has already been moved into `Unit`.

sim/core/stats/stats.go
```diff
@@ -332,6 +332,9 @@ type PseudoStats struct {
 	DamageDealtMultiplier       float64            // All damage
 	SchoolDamageDealtMultiplier [SchoolLen]float64 // For specific spell schools (arcane, fire, shadow, etc).
 
+	// Apply Aura: Mod Melee Critical Damage % (Arcane, Fire, Frost, Holy, Nature, Physical, Shadow)
+	SchoolCritMultiplier [SchoolLen]float64 // For specific spell schools (arcane, fire, shadow, etc).
```
jimmyt857, line 336: I don't see this used anywhere

rosenrusinov: Hodir Stormpower buff the comment is how the wow db has it listed in the aura effects

sim/core/unit.go
rosenrusinov: Did not see that will use it instead 👍

sim/core/stats/stats.go
rosenrusinov: Just noticed The hodir_ai file is not shown by default from github due to its size so probably missed it

PR #2606, opened 2023-02-06
Add SoV Buff and HoR to Ret Custom and Standard rotation

Foknetics:
Implementing the proposed [Retribution Buffs](https://us.forums.blizzard.com/en/wow/t/upcoming-adjustments-to-retribution-paladins/1511714). Still will need the actual Glyph ID once added for proper tooltip support.

### DIFF
--- a/sim/paladin/hand_of_reckoning.go
+++ b/sim/paladin/hand_of_reckoning.go
@@ -4,9 +4,14 @@ import (
 	"time"
 
 	"github.com/wowsims/wotlk/sim/core"
+	"github.com/wowsims/wotlk/sim/core/proto"
 )
 
 func (paladin *Paladin) registerHandOfReckoningSpell() {
+	if !paladin.HasMajorGlyph(proto.PaladinMajorGlyph_GlyphOfReckoning) {
+		return
+	}
+
 	paladin.HandOfReckoning = paladin.RegisterSpell(core.SpellConfig{
 		ActionID:    core.ActionID{SpellID: 62124},
 		SpellSchool: core.SpellSchoolHoly,
--- a/sim/paladin/hand_of_reckoning.go
+++ b/sim/paladin/hand_of_reckoning.go
@@ -4,7 +4,6 @@ import (
 	"time"
 
 	"github.com/wowsims/wotlk/sim/core"
-	"github.com/wowsims/wotlk/sim/core/proto"
 )
 
 func (paladin *Paladin) registerHandOfReckoningSpell() {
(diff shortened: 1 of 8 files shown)


Foknetics:
This does not cover using HoR in the cast sequence rotation type. HoR is used on CD if the glyph is selected.

sim/paladin/hand_of_reckoning.go
```diff
@@ -0,0 +1,47 @@
...
+		CritMultiplier:           paladin.SpellCritMultiplier(),
+
+		ApplyEffects: func(sim *core.Simulation, target *core.Unit, spell *core.Spell) {
+			baseDamage := 1 +
+				.5*spell.MeleeAttackPower()
+
+			bonusHit := core.TernaryFloat64(
+				paladin.HasMajorGlyph(proto.PaladinMajorGlyph_GlyphOfRighteousDefense),
+				8*core.SpellHitRatingPerHitChance,
+				0)
```
rosenrusinov, line 40: cache this value outside of the ApplyEffects function

sim/paladin/hand_of_reckoning.go
```diff
@@ -0,0 +1,47 @@
...
+
+			spell.BonusHitRating += bonusHit
+			spell.CalcAndDealDamage(sim, target, baseDamage, spell.OutcomeMagicHitAndCrit)
+			spell.BonusHitRating -= bonusHit
```
rosenrusinov, line 44: Do you need to apply and remove this bonus hit for any reason? Can this spell be cast without this hit? If not you should set this up once in the SpellConfig and just deal damage here

Foknetics: The only reason for it to be added is if the glyph is selected. The spell can be used fine without the glyph. This was mostly following the flow exorcism used for applying bonus crit rating to undead mobs which felt similar in a conditional rating change

sim/paladin/sov.go
```diff
@@ -182,8 +182,8 @@ func (paladin *Paladin) registerSealOfVengeanceSpellAndAura() {
 				}
 			}
 
-			// Only white hits and HotR can trigger this. (SoV dot)
-			if spell.ProcMask.Matches(core.ProcMaskMeleeWhiteHit) || spell.SpellID == paladin.HammerOfTheRighteous.SpellID {
+			// Only white hits, HotR, CS, and DS can trigger this. (SoV dot)
+			if spell.ProcMask.Matches(core.ProcMaskMeleeWhiteHit) || spell.SpellID == paladin.HammerOfTheRighteous.SpellID || spell.SpellID == paladin.CrusaderStrike.SpellID || spell.SpellID == paladin.DivineStorm.SpellID {
```
jimmyt857, line 186: We can compare spell pointers directly instead of `SpellID`s, e.g. `spell == paladin.CrusaderStrike`

sim/paladin/hand_of_reckoning.go
```diff
@@ -0,0 +1,39 @@
+package paladin
+
+import (
+	"time"
+
+	"github.com/wowsims/wotlk/sim/core"
+)
+
+func (paladin *Paladin) registerHandOfReckoningSpell() {
```
jimmyt857, line 9: Can we start this function with a check like so:
```
if paladin.HasMajorGlyph(proto.PaladinMajorGlyph_GlyphOfReckoning) {
  return
}
```

And then later just check if `paladin.HandOfReckoning == nil` to determine the presence of the spell. This will play more nicely with future APL changes, and allow the `ret.HandOfReckoning` boolean to be removed.

PR #2609, opened 2023-02-06
Update ferocious_bite.go

W1ldish:


jarveson:
this isnt correct, it looks like your changing values to 'rank 6' bite instead of 'rank 8'

PR #2717, opened 2023-02-24
dirty deeds

vigo2:
[rogue] make dirty deeds work


catszeid:
Looks good! Thanks for adding that in.

catszeid (approved):

PR #2726, opened 2023-02-27
Do not give rune grace for pre-pull casts

lime-green:
Doesn't really affect normal rotation

For desync rotation:

Before:
image

After:
image

Army was causing runes to go on rune grace in the opener and syncing up the runes for the desync rotation, but runes used before combat should not incur rune grace


sim/core/runic_power.go
```diff
@@ -257,9 +257,16 @@ func (rp *RunicPowerBar) SpentDeathRuneReadyAt() time.Duration {
 	return readyAt
 }
 
-func (rp *RunicPowerBar) CurrentRuneGrace(sim *Simulation, slot int32) time.Duration {
-	if rp.runeMeta[slot].lastRegenTime < sim.CurrentTime {
-		return time.Millisecond*2500 - MinDuration(2500*time.Millisecond, sim.CurrentTime-rp.runeMeta[slot].lastRegenTime)
+func (rp *RunicPowerBar) RuneGraceRemaining(sim *Simulation, slot int32) time.Duration {
```
lime-green, line 260: renamed since this was a misleading name

rosenrusinov (approved):

PR #2730, opened 2023-03-01
More Subtlety changes

vigo2:
[core] add sim.RandomExpFloat(), essentially a wrapper around rand.ExpFloat64()

[rogue] only enable hemoAuras for the raid sim (if there are >= 2 players) 

[rogue] introduce honor_of_thieves_crit_rate option, defaulting to 400. it contains the number of ability crits generated by other party members within 100 seconds 

[rogue] make subtlety rotation depend on HaT > 0 instead of having Shadowstep (it's the most influental talent) 

[rogue] support Glyph of Shadow Dance, and improve Shadow Dance's ShouldActivate() 

[rogue] HaT procs are now generated from the rogue's own ability crits, and other party members, modelled as a Poisson distribution 

[rogue] 2t10 now always grants 15 energy on TotT usage

### DIFF
--- a/sim/rogue/talents.go
+++ b/sim/rogue/talents.go
@@ -605,10 +605,27 @@ func (rogue *Rogue) registerHonorAmongThieves() {
 		return
 	}
 
-	// FIXME: Implement a flow of combo points that simulates the party/raid crit procs
+	// In an ideal party, you'd probably get up to 6 ability crits/s (Rate = 600).
+	//  Survival Hunters, Enhancement Shamans, and Assassination Rogues are particularly good.
+	if rogue.Options.HonorOfThievesCritRate <= 0 {
+		rogue.Options.HonorOfThievesCritRate = 400
+	}
+
 	procChance := []float64{0, 0.33, 0.66, 1}[rogue.Talents.HonorAmongThieves]
 	comboMetrics := rogue.NewComboPointMetrics(core.ActionID{SpellID: 51701})
 	honorAmongThievesID := core.ActionID{SpellID: 51701}
+
+	icd := core.Cooldown{
+		Timer:    rogue.NewTimer(),
+		Duration: time.Second,
+	}
+
+	maybeProc := func(sim *core.Simulation) {
+		if icd.IsReady(sim) && sim.Proc(procChance, "honor of thieves") {
+			rogue.AddComboPoints(sim, 1, comboMetrics)
+			icd.Use(sim)
+		}
+	}
 
 	rogue.HonorAmongThieves = rogue.RegisterAura(core.Aura{
 		Label:    "Honor Among Thieves",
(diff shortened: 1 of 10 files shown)


sim/rogue/talents.go
```diff
@@ -605,27 +605,56 @@ func (rogue *Rogue) registerHonorAmongThieves() {
...
 	procChance := []float64{0, 0.33, 0.66, 1}[rogue.Talents.HonorAmongThieves]
 	comboMetrics := rogue.NewComboPointMetrics(core.ActionID{SpellID: 51701})
 	honorAmongThievesID := core.ActionID{SpellID: 51701}
 
+	icd := core.Cooldown{
+		Timer:    rogue.NewTimer(),
+		Duration: time.Second,
+	}
+
+	maybeProc := func(sim *core.Simulation) {
```
catszeid, line 623: The current model for the proc is that the ICD is triggered, even on a failed roll that gives no CP. It does not appear to do so in this snippet

sim/rogue/talents.go
```diff
@@ -605,27 +605,56 @@ func (rogue *Rogue) registerHonorAmongThieves() {
...
+			pa.OnAction = func(sim *core.Simulation) {
+				maybeProc(sim)
+				pa.NextActionAt = sim.CurrentTime + time.Duration(sim.RandomExpFloat("next party crit")*rateToDuration)
+				sim.AddPendingAction(pa)
+			}
+			pa.NextActionAt = sim.CurrentTime + time.Duration(sim.RandomExpFloat("next party crit")*rateToDuration)
+			sim.AddPendingAction(pa)
+		},
+		OnSpellHitDealt: func(_ *core.Aura, sim *core.Simulation, spell *core.Spell, result *core.SpellResult) {
+			if result.DidCrit() && !spell.ProcMask.Matches(core.ProcMaskMeleeMHAuto|core.ProcMaskMeleeOHAuto|core.ProcMaskRangedAuto) {
```
catszeid, line 650: Honor Among Thieves only triggers on ability criticals, so I think you may need to modify this to make sure generic spell hits (Poisons, trinket procs) do not trigger it.

catszeid (changes requested):
Looked over the code and looks good except a few small things.

sim/rogue/talents.go
vigo2: It triggers on any crit that's not an auto attack, including poisons and trinket procs. https://www.wowhead.com/wotlk/spell=51701/honor-among-thieves#comments:id=893977 has a (non-exhaustive) list, and I've verified most of those some months ago.

sim/rogue/talents.go
vigo2: It's highly unlikely that ICD is invoked without an actual proc, since nothing else works this way. I'll test on PTR, but that's very tedious since CPs aren't logged.

catszeid (approved):

PR #2767, opened 2023-03-12
flare of the heavens should only proc on harmful spells

lime-green:


lime-green:
this doesn't actually do anything since Harmful flag is only used with the `CallbackOnSpellHitDealt` callback. `ProcMaskSpellDamage` will already be checked and that should be sufficient!

PR #2775, opened 2023-03-13
improve frost sub blood BS logic

lime-green:
Improves rotation to not cast frost strike before blood strike, there's no point since we optionally cast a KM FS before blood strike, and will cast a FS after the blood strike if we can before obliterate.

On 3 minutes it's a 34 dps increase

### DIFF
--- a/sim/deathknight/dps/rotation_frost.go
+++ b/sim/deathknight/dps/rotation_frost.go
@@ -9,7 +9,6 @@ import (
 
 type FrostRotation struct {
 	oblitCount int32
-	uaCycle    bool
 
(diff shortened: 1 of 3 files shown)


sim/deathknight/dps/rotation_frost.go
```diff
@@ -9,7 +9,6 @@ import (
 
 type FrostRotation struct {
 	oblitCount int32
-	uaCycle    bool
```
lime-green, line 12: this is only ever set to false, does nothing

rosenrusinov (approved):

PR #2776, opened 2023-03-13
DK Bcb scaling & imp unh presence updates

rosenrusinov:
- Fix bcb scaling to work with external CF/EF as it does on live
- Stop rune holding for Dnd when improved unholy presence is talented


lime-green:
nice, frost UH lgtm. Did you test the BCB scaling with external EB or is there a log?

rosenrusinov:
> nice, frost UH lgtm. Did you test the BCB scaling with external EP or is there a log?

I dont have logs atm but i have 2 accounts with an 80 dk in both and it was very easy yo test this on a dummy as the damage difference is significantly big to notice it easy

PR #2781, opened 2023-03-14
bryntroll cant crit

lime-green:
This can't crit according to multiple comments from https://www.wowhead.com/wotlk/item=50415/bryntroll-the-bone-arbiter#comments


lime-green:
vigo should be doing this

PR #2789, opened 2023-03-16
Rogue rotation updates

vigo2:
[rogue] no longer delay DPS cooldowns waiting for armor debuffs
[rogue] revert to casting all poison, so they appear no the timeline 
[rogue] fix Shiv being able to crit (and be dodged, blocked or parried) 
[rogue] update assassination rotation to use generic (non-configurable) pooling, properly use Cold Blood, end-of-fight handling, and (sparingly) use Rupture in high-energy return situations 
[rogue] update subtlety rotation to Build faster towards Rupture, end-of-fight handling, and more responsive handling of "I'm sitting at 4 CP and just wait for HAT" situations 
[rogue] fixed some wrong talent icons (Improved Eviscerate, Cut to the Chase) 
[rogue] added "HonorOfThievesCritRate" input, and correctly updated the UI when changing between Specs 
[rogue] changed some rotation, option, and talent defaults, and added gear presets for "Hemo Sub"


b-r-a-n (commented):
Looks like all positive changes to me. I'm less familiar with the subtlety stuff but that looks good to me as well.

b-r-a-n (approved):

PR #2798, opened 2023-03-19
Mutilate Fix (and only that!)

vigo2:
[rogue] The Mutilate "parent" spell triggered Focused Attacks whenever any of the damage spells crit, returning up to 6 instead of only 4 energy. Each damage spell now counts as a builder for Seal Fate, which has gained an ICD so Mutilate doesn't grant up to 4 CPs

[rogue] Seal Fate now has a proper ICD (500 ms, as used in-game as well) 

[rogue] Deadly Poison now properly snapshots BaseDamage and AttackerMutliplier on each refresh. The DPS difference is negligible, as expected


catszeid (approved):
Looks good. Nothing jumped out to me as needing fixed

PR #2867, opened 2023-04-05
Implement first version of the BulkSim ("sim inventory") UI

zku:
Implements an initial version of a UI for the BulkSim feature.

### What does BulkSim mean here?

Bulk sim here means "sim my inventory" - a feature that has been requested a few times on Discord. The motivation behind this is simple: **I have a bunch of equipped gear, and I have a bunch of items in my inventory that may or may not be BIS depending on what other items I have available**.

For example, a combination of items in my bags might be an improvement over my currently equipped gear, but individually they are all downgrades due to loss of e.g. a tier set bonus. Some item might only be useable if I also change another item to remain above hit cap, and so on.

Simulating the performance all these potential item combinations is tedious in the single-gear user interface: people need to manually select the extra items they have (including gems and enchants), and then need to manually sim every possible item combo (the number grows exponentially).

Retail simulators already support this "sim my bags" or "sim my vault" feature. It is essentially a combination of equipped items plus items in the bags.

*Note that there might be different ideas of what a "bulk sim" is, such as a free form selection of possible item combos. This is not the goal here. The goal here is to serve the most practical use case of simming all the items I have available (equipped + bags) in all possible combinations, not to sim hypothetical combinations.

#### Web related changes

Needles to say, I suck at this. First time touching TypeScript and I am not a fan.

* Adds a new `bulk_tab.ts` file to implement a new top-level tab to specify the bulk configuration. Items can be imported via the existing (adjusted) exporter WoW AddOn.
* Refactors (butchers?) the `ItemPicker` into two separate components: one that is a view-only component responsible for the rendering of items, and the `ItemPicker` that has on-click actions that modify the player's equipment. The bulk items cannot be modified (the point is to sim your bags, not hypothetical items) and just use the view-only `ItemRenderer`. The reason this is somewhat required is because the `ItemPicker` is tightly coupled to an `ItemSlot`, whereas for bag simming it is entirely possible that multiple items for the same slot exist in someone's bags.
* Minor adjustments in other files, e.g. to return the proper item slots for a given item, and to update the plumbing code around propagating sim item databases.
* The bulk tab is "hidden" behind the experimental flag, i.e. people need to opt in to use this by checking the experimental checkbox in their settings.

#### WoW addon changes

* Adds a new button to export bag data.
* Fixes the broken import order of the addon libraries (this works for most people right now because anyone using this AddOn probably also has other AddOns installed and has correct libraries already loaded)
* Fixes broken APIs in one of the libs (this works for most people right now because anyone using this AddOn probably also has other AddOns installed and has correct libraries already loaded)

#### Core changes

* Adds `bulksim.go` and `bulksim_test.go` next to `sim.go` to implement the bulk simming. The reason this doesn't use the existing CLI code is because of some desired changes in output (consistent iterations for all sims) and because I couldn't initially identify why it wasn't working properly, and only later identified some bug(s). See Discord for more context.

Small video of this in action can be seen here: https://discord.com/channels/891730968493305867/1092705282452365314/1092891569260085411


package-lock.json
lologarithm, line 1: We don't commit the package-lock.json - just remove it here for now (I thought we had it in the .gitignore...)

sim/core/sim.go
```diff
@@ -42,7 +43,27 @@ type Simulation struct {
 	executePhaseCallbacks []func(*Simulation, int) // 2nd parameter is 35 for 35%, 25 for 25% and 20 for 20%
 }
 
-func RunSim(rsr *proto.RaidSimRequest, progress chan *proto.ProgressMetrics) (result *proto.RaidSimResult) {
+func RunSim(rsr *proto.RaidSimRequest, progress chan *proto.ProgressMetrics) *proto.RaidSimResult {
+	// TODO(Riotdog-GehennasEU): expose this as a separate API?
```
lologarithm, line 47: I actually would recommend this! I can help do this

PR #2882, opened 2023-04-08
Added sim all button to item pickers

lologarithm:


sim/core/major_cooldown.go
```diff
@@ -144,7 +144,12 @@ func (mcd *MajorCooldown) tryActivateHelper(sim *Simulation, character *Characte
 	}
 
 	if shouldActivate {
-		mcd.Spell.Cast(sim, character.CurrentTarget)
+		if mcd.Spell.Flags.Matches(SpellFlagHelpful) {
```
lologarithm, line 147: turns out we had an outstanding bug where helpful trinkets actually targeted the enemy not ourself

Issue #1512, opened 2022-10-29
Bulk Sims


jimmyt857:
Making a master issue to centralize the conversation around this. The goal is to be able to configure combinations of settings and then run all of them. A simple example would be "find the best item for this slot" or "show me how this item performs with all of my different rotation setups". We can go further though, and allow multidimensional sims and produce great visuals showing the results.

jimmyt857:
Copied from a discord discussion on this:

@Pokpoong:
Unfortunately the issues are deeper than that.

Let's start with the user need, which is basically, "tell me which items are bis / near bis for this slot". This is a perfectly valid need that we should try to fill. The problem is that stats are way more complicated now, to the point where it is not possible to choose an item in isolation.

To give a simple example: pick any slot and check out the top 3-5 items in that slot for your character. I gaurantee some will have hit/expertise stats, and some won't. Already this presents the problem: to get the full value out of an item with extra hit/expertise (or any stat with a cap), you need to choose alternatives for different items that trade the overcapped hit for other uncapped stats.

Note that this is still ignoring the gem issue.

Guybrush-style UI worked back in classic, BECAUSE it was classic. The set of relevant items in the entire game could fit on a small fraction of the screen, and stats were simple enough that there was almost always exactly 1 choice for BiS. Replicating the same feature now results in something which doesn't solve the user's need. So we need to come up with something smarter, which solves the actual need.

jimmyt857:
Copied from a discord discussion on this:

@lologarithm:
I think there is two different kinds of bulk sims, but I think they can have a single UI -- 

1. Give me the best combination of items out of a pile of different items (like say 1-3 items in each slot)  This could also be used for comparing a couple slots of different items at a time (like say comparing set bonuses vs better item with no set bonus)

2. Give me the best item for this slot (could be like 30-50 items in the slot).  This is best for something like trinkets/relics/ etc where they have very little in the way of EPs but would be useful to roughly order them.
well, there is also the "just sim every item and tell me the best" but I think that simulation would take years to complete and isn't worth aiming for

jimmyt857:
Copied from a discord discussion on this:

@Pokpoong:
I think we should be thinking more generally
Simon has a lot of great charts in his spreadsheets, here's a screenshot of one:
bulk sims should let you sim across all sorts of dimensions, with gear being just 1 option for dimensions

we basically need a proto for "SettingsDimension" with lots of different sub-options; different types of gear dimensions, a rotation dimension, encounter, talents, etc
[...]

PR #2883, opened 2023-04-08
Render BulkUI items in a grid, prepare changes to make items removable / changeable

zku:
* Render the items in a grid (for now they still display their names/enchants etc, but we can think about removing that as well and just show icons)
* Adapt the grid based on the screen size (3 -> 2 -> 1 columns)
* Create a BulkItemPicker that will be the base to delete items from the bulk or to change them


lologarithm:
Continues: https://github.com/wowsims/wotlk/issues/1512

PR #2885, opened 2023-04-08
bump node and use faster watcher

lime-green:


makefile
```diff
@@ -164,8 +164,8 @@ endif
 
 rundevserver: air devserver
 ifeq ($(WATCH), 1)
-	ulimit -n 10240 && air -tmp_dir "/tmp" -build.include_ext "go,ts,js,html" -build.bin "true" -build.cmd "make" -build.exclude_dir "dist,node_modules,sim,tools" &
-	ulimit -n 10240 && air -tmp_dir "/tmp" -build.args_bin "--usefs=true --launch=false" -build.bin "./wowsimwotlk" -build.cmd "make devserver" -build.exclude_dir "assets,dist,node_modules,ui,tools"
+	npx vite build -m development --watch &
```
lologarithm, line 167: does CTRL+C kill both processes or do you have to kill the process directly when you do this?

lime-green: ctrl-c kills it!

PR #2892, opened 2023-04-09
ItemPicker and BulkUI render updates

zku:
* Reuses the existing `SelectorModal` to render enchant/gem selection for bulk items
* Adds a new button to destroy the item (remove from bulk)
* Fixes bug in the existing `SelectorModal` gem tab where removing a gem would not change the gem icon to empty
* Renames the "Unequip Item" button depending on the tab: remains the same for the items tab, but is renamed to "Remove Enchant" and "Remove Gem" in the enchant and gem tabs, respectively. Before the button would always be "Unequip Item" even though it acted on the enchant or gem.

### DIFF
--- a/ui/core/components/gear_picker.ts
+++ b/ui/core/components/gear_picker.ts
@@ -450,7 +450,7 @@ interface GearData {
 	changeEvent: TypedEvent<any>,
 }
 
-enum SelectorModalTabs {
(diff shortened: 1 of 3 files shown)


lologarithm:
Continues: https://github.com/wowsims/wotlk/issues/1512

PR #2898, opened 2023-04-10
Check if a cast can be completed before sending ShouldCast

catszeid:
Add checks to the generic rogue rotation to ensure that addition cast requirements are satisfied before ShouldCast is returned and causes an unexpected cast failure.

### DIFF
--- a/sim/rogue/rotation_generic.go
+++ b/sim/rogue/rotation_generic.go
@@ -288,6 +288,7 @@ func (x *rotation_generic) run(sim *core.Simulation, rogue *Rogue) {
 		} else {
 			panic("Unexpected builder cast failure")
 		}
+
 	case ShouldCast:
 		spell := prio.GetSpell(rogue, rogue.ComboPoints())
 		if spell == nil || spell.Cast(sim, rogue.CurrentTarget) {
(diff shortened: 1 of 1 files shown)


sim/rogue/rotation_generic.go
```diff
@@ -395,7 +396,7 @@ func (x *rotation_generic) shouldCastNextRotationItem(sim *core.Simulation, rogu
 		if tte > item.MinimumBuildDuration {
 			// Find the first lower prio item that can be cast and use it
 			for lpi, lowerPrio := range x.priorityItems[item.PrioIndex+1:] {
-				if comboPoints > lowerPrio.MinimumComboPoints && currentEnergy > lowerPrio.EnergyCost && lowerPrio.MaxCasts == 0 {
+				if comboPoints > lowerPrio.MinimumComboPoints && currentEnergy > lowerPrio.EnergyCost && lowerPrio.MaxCasts == 0 && prio.GetSpell(rogue, comboPoints).CanCast(sim, rogue.CurrentTarget) {
```
vigo2, line 399: should be "lowerPrio" instead

sim/rogue/rotation_generic.go
```diff
@@ -374,7 +375,7 @@ func (x *rotation_generic) shouldCastNextRotationItem(sim *core.Simulation, rogu
 
 	// Expires before next GCD
 	if tte <= timeUntilNextGCD {
-		if comboPoints >= prio.MinimumComboPoints && currentEnergy >= (prio.EnergyCost+prio.PoolAmount) {
+		if comboPoints >= prio.MinimumComboPoints && currentEnergy >= (prio.EnergyCost+prio.PoolAmount) && prio.GetSpell(rogue, comboPoints).CanCast(sim, rogue.CurrentTarget) {
```
vigo2, line 378: prio.GetSpell() returns `nil` for the MCD enabler, and spell.CanCast() returns `false` in this case - which might end up in an endless `return ShouldWait` (here or elsewhere)

vigo2 (changes requested):
You can change this PR (~add a local CanCast() also checking for `nil`, or per PrioItem CanCast()s), or simply wait for my upcoming PR ;) 

Even with a "proper" CanCast(), the rotation can very likely get stuck waiting indefinitely, so the crashes could just as well be addressed by adding `rogue.DoNothing()` instead of the current panics.

PR #2900, opened 2023-04-10
Restrict dual wielding Polearm and Staff

TheGroxEmpire:


ui/core/proto_utils/utils.ts
```diff
@@ -1810,7 +1810,9 @@ export function validWeaponCombo(mainHand: Item | null | undefined, offHand: Ite
 		return true;
 	}
 
-	if (mainHand.handType == HandType.HandTypeTwoHand && !canDW2h) {
+	if (mainHand.handType == HandType.HandTypeTwoHand && !canDW2h ||
+		mainHand.handType == HandType.HandTypeTwoHand && 
+		(mainHand.weaponType == WeaponType.WeaponTypePolearm || mainHand.weaponType == WeaponType.WeaponTypeStaff)) {
```
vigo2, line 1815: nit: I'd prefer splitting this into (at least) two more readable `if` clauses, one for `canDW2h` and one for `!canDW2h`

TheGroxEmpire: What do you think of 024873dc9?

vigo2: much better :)

PR #2902, opened 2023-04-10
PR to compute weighted EP from a mix of DPS/TPS/DTPS

merijn:
Basically, we assign a ratio to each EP column and update the current EP with a weighted sum of those columns. Needs two things to be ready for the main sim:

1) The UI looks terrible currently, someone more comfortable with the CSS/front-end stuff can probably make it look better,

2) should probably default to hiding the number cells and compute button for DPS specs, who will only be interested in DPS EP anyway

### DIFF
--- a/ui/core/components/stat_weights_action.ts
+++ b/ui/core/components/stat_weights_action.ts
@@ -115,15 +115,15 @@ class EpWeightsMenu extends BaseModal {
 									<i class="fa fa-copy"></i>
 								</a>
 							</th>
-							<th>
+							<th style="text-align: center;">
 								<span>Current EP</span>
 								<a href="javascript:void(0)" role="button" class="col-action">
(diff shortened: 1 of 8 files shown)


ui/core/components/stat_weights_action.ts
```diff
@@ -122,6 +122,31 @@ class EpWeightsMenu extends BaseModal {
 								</a>
 							</th>
 						</tr>
+						<tr>
+							<th>Ratio</th>
+							<td class="damage-metrics type-ratio type-weight">
+							</td>
+							<td class="damage-metrics type-ratio type-ep">
```
lologarithm, line 129: add to _stat_weights_action.scss 
```.results-ep-table-container .type-ratio {
  text-align: right;
}```

To make these boxes align with the text columns under them

merijn:
@lologarithm imo, this is ready to merge.

PR #2906, opened 2023-04-10
Batch Auto Gem and Auto Enchant

lologarithm:
Adds options to auto-gem and auto-enchant items for batch simulator.


JordanMLewis:
Good idea! Looks like a good change.

PR #2961, opened 2023-04-18
update stats proto to keep go and proto in sync, regenerate db

zku:


zku:
This should be the actual fix for #2960 which should make the encounter dropdown visuals work (it always worked and set the correct encounter, but the UI would always fall back to "Custom" instead).

PR #2968, opened 2023-04-18
^_^

caccavale:


lologarithm:
implemented in https://github.com/wowsims/wotlk/pull/3129

PR #2973, opened 2023-04-19
Remove unnecessary function literal creations

zku:
Remove some unnecessary function literal creations (major contributor to small GC allocations -> sweeps). A lot of time is spent allocating and cleaning up after all the new function literals being created, and some of these don't really have to exist, i.e. a pattern like

```OnAction = func(sim *Simulation) { Foo(sim) }```

could just be expressed as

```OnAction = Foo```

without the additional allocation (this is already used in most places luckily). Over 100k iterations, it seems like we spend most of our time creating and re-scheduling `PendingAction`s, which all create these unnecessary function literals repeatedly.

Using the raid bench with 1000 iterations and 300s duration over 10 repetitions yields a ~25% allocs/op reduction with minor performance gains due to GC activity reductions (~5% faster). Of course this will be very machine specific, but reducing small unnecessary allocations should be an improvement under most circumstances.

Lastly, also changes the unit test to bind to localhost (vs. all interfaces) to prevent Windows UAC prompts when running the unit tests.

For reference:

```
goos: windows
goarch: amd64
pkg: github.com/wowsims/wotlk/sim
cpu: AMD Ryzen 7 5800X 8-Core Processor
```

### DIFF
--- a/sim/core/gcd.go
+++ b/sim/core/gcd.go
@@ -15,9 +15,7 @@ func (unit *Unit) newHardcastAction(sim *Simulation) {
 	if unit.hardcastAction == nil {
 		pa := &PendingAction{
 			NextActionAt: unit.Hardcast.Expires,
-			OnAction: func(sim *Simulation) {
-				// Don't need to do anything, the Advance() call will take care of the hardcast.
-			},
+			// No action, the Advance() call will take care of the hardcast.
 		}
 		unit.hardcastAction = pa
 	} else {
--- a/sim/core/gcd.go
+++ b/sim/core/gcd.go
@@ -15,7 +15,9 @@ func (unit *Unit) newHardcastAction(sim *Simulation) {
 	if unit.hardcastAction == nil {
 		pa := &PendingAction{
 			NextActionAt: unit.Hardcast.Expires,
-			// No action, the Advance() call will take care of the hardcast.
+			OnAction: func(sim *Simulation) {
+				// Don't need to do anything, the Advance() call will take care of the hardcast.
+			},
 		}
 		unit.hardcastAction = pa
 	} else {
--- a/sim/core/periodic_action.go
+++ b/sim/core/periodic_action.go
@@ -52,6 +52,7 @@ func NewPeriodicAction(sim *Simulation, options PeriodicActionOptions) *PendingA
 	pa := &PendingAction{
 		NextActionAt: sim.CurrentTime + options.Period,
 		Priority:     options.Priority,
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: sim/core/periodic_action.go [hunk 1]
```diff
@@ -15,21 +15,12 @@ type DelayedActionOptions struct {
 }
 
 func NewDelayedAction(sim *Simulation, options DelayedActionOptions) *PendingAction {
-	pa := &PendingAction{
```
vigo2, line 18: it's simpler to panic() if `options.OnAction` is nil here...

sim/core/periodic_action.go
```diff
@@ -61,12 +52,15 @@ func NewPeriodicAction(sim *Simulation, options PeriodicActionOptions) *PendingA
 	pa := &PendingAction{
```
vigo2, line 52: ... and here as well. This allows to keep the implicit "PendingAction.OnAction() must not be nil" invariant elsewhere, and makes it more explicit.
The code changes become very minimalistic then, too :)

sim/core/gcd.go
```diff
@@ -15,9 +15,7 @@ func (unit *Unit) newHardcastAction(sim *Simulation) {
 	if unit.hardcastAction == nil {
 		pa := &PendingAction{
 			NextActionAt: unit.Hardcast.Expires,
-			OnAction: func(sim *Simulation) {
-				// Don't need to do anything, the Advance() call will take care of the hardcast.
-			},
+			// No action, the Advance() call will take care of the hardcast.
```
vigo2, line 18: You'd have to leave this "as is" then, which isn't much of a problem, as this case is very rarely taken. The main culprits are NewDelayedAction() and NewPeriodicAction().
Alternatively, core could add a `func DoNothing(*Simulation) {}`, but that's probably overkill.

vigo2:
Otherwise, LGTM. Thanks for doing this!

vigo2 (changes requested):

sim/core/periodic_action.go
zku: Done!

sim/core/periodic_action.go
zku: Done!

sim/core/gcd.go
zku: Reverted.

lologarithm (approved):

zku:
Can you please have another look @vigo2 ?

vigo2 (approved):

Issue #2983, opened 2023-04-21
support glyph ids


Nordwald:
**Is your feature request related to a problem? Please describe.**
I'm using a german client. Whenever I use the WoWSims Exporter Addon, there is no information about glyphs contained (`"glyphs":{"major":[],"minor":[]}`. Everytime I import my character, I have to manually adjust the glyphs again.

After digging into the issue, I found the problem at line 75 in `WowSimsExporter.lua`:
```lua
			local name, _ = string.match(link,"Glyph of .+]")
			if(name) then
				local formattedName = name:gsub('%]', '')
```

In German, the Glyph of Holy Light is named `Glyphe 'Heiliges Licht'`.
However, as far as I can tell the AddOn and the website use the string names of glyphs. So in order to support another lanugage, one would need to translate every glyph name. 

**Describe the solution you'd like**
In order to fix this, you can support glyphs by id analogous to how ids are used for enchants and gems already. This would allow for easy support of other languages, since we don't even need to extract the name at all.

**Describe alternatives you've considered**
I've considered two other possible solutions:
- Support trnaslation support for glyüphs
- Give me an option not to overwrite glyphs with every import

Nordwald:
I've been digging deeper, and it seems this feature request could require very little change.

`/ui/core/components/importers.ts`
```lua
		const glyphsConfig = classGlyphsConfig[charClass];
		const majorGlyphIDs = importJson['glyphs']['major'] as Array<string>).map(glyphName => glyphNameToID(glyphName, glyphsConfig.majorGlyphs));
		const minorGlyphIDs = (importJson['glyphs']['minor'] as Array<string>).map(glyphName => glyphNameToID(glyphName, glyphsConfig.minorGlyphs));
```
Here we can remove the code getting a glyph's id from its name, since we can pass the id in the export string directly.

```lua
		local enabled, glyphType, glyphTooltipIndex, glyphSpellID, icon = GetGlyphSocketInfo(t);
		local link = GetGlyphLink(t);

		if(enabled) then	
			
			local name, _ = string.match(link,"Glyph of .+]")
			if(name) then
				local formattedName = name:gsub('%]', '')

				if(glyphType == 1 ) then-- major
					table.insert(major, formattedName)
				else if(glyphType == 2 ) then -- minor
					table.insert(minor, formattedName)
				end
			end

		end
````
In this file, we could simplfy the loop with id export - I'd propose something like this:
```lua
[...]

PR #2985, opened 2023-04-21
Glyph fix

Nordwald:


zku:
Thanks for this! Per https://github.com/wowsims/wotlk/issues/2983#issuecomment-1518395931 I'll close this PR in favor of the other one since this one would not work as intended (spell IDs vs. item IDs).

PR #3036, opened 2023-05-04
do not populate ActionID for proc trigger auras

zku:
Instead refer to the proc aura that the synthetic trigger aura refers to. This way no synthetic trigger aura will ever be transmitted through the metrics, and the trinket desync functionality can still find the relevant synthetic aura to apply the desync to.

### DIFF
--- a/sim/core/aura.go
+++ b/sim/core/aura.go
@@ -44,7 +44,8 @@ type Aura struct {
 	// For easily grouping auras.
 	Tag string
 
-	ActionID ActionID // If set, metrics will be tracked for this aura.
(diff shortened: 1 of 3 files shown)


zku:
Historically, the code here:

```
func MakeProcTriggerAura(unit *Unit, config ProcTrigger) *Aura {
	aura := Aura{
		Label:    config.Name,
		ActionID: config.ActionID,
```

would set the ActionID based on the config's action ID. Some configs used to set it, and some did not. With this change, we ensure that the ActionID will never be set for such synthetic auras. This should fix the timeline/results rendering issue where synthetic auras were also popping up sometimes.

PR #3159, opened 2023-06-15
Allow Combat Rotation to cast Ghostly Strike

where-fore:
For Science!

Pretty hacky code, added upon request. Has almost no practical application, but at least now we can see that it's only an 1182 dps loss to spec down into sub and grab Ghostly Strike and glyph it.


where-fore:
Oh dear I combined this with my other pull.. please hold.

PR #3162, opened 2023-06-15
Mage p3 presets

lycaste:
Mage p3 presets

### DIFF
--- a/ui/mage/presets.ts
+++ b/ui/mage/presets.ts
@@ -62,6 +62,21 @@ export const FireTalents = {
 		}),
 	}),
 };
+
+export const Phase3FireTalents = {
+	name: 'Fire Phase 3',
+	data: SavedTalents.create({
+		talentsString: '23002303310003-0055030012303330053120300351',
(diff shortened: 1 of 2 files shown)


lycaste:
@jimmyt857 workflow requires approval from a maintainer

ui/mage/presets.ts
```diff
@@ -1134,4 +1149,245 @@ export const FFB_P2_PRESET = {
 		  ]
 		}
 	  ]}`),
+};
+
+
+export const FIRE_P3_PRESET_HORDE = {
+	name: 'Fire P3 Preset Horde',
+	enableWhen: (player: Player<Spec.SpecBalanceDruid>) => player.getFaction() == Faction.Horde,
```
jimmyt857, line 1157: This condition needs to also check for the player spec, e.g. `player.getTalentTree() == 1`. We don't want to show fire presets when the user has arcane settings, for example.

Same comment for the alliance preset.

lycaste: ok. fixed it

Issue #2173, opened 2022-12-15
APL-like rotation input


jimmyt857:
Design a fully-customizable way (similar to APL in simc) which allows complete control over the character's rotation. Ideally, this system will be able to replace most or all existing rotations.

Here's a good reference for APL: https://github.com/simulationcraft/simc/wiki/ActionLists

In WoWSims, we probably will use a proto schema for this rather than raw strings. The `CharacterStats` API call could return all available spells / auras for each unit, allowing the UI to give the user dropdown lists for all the options rather than forcing them to memorize spell IDs.

jimmyt857:
An important prerequisite for this will be to refactor the code for every spec to remove all assumptions of `CurrentTarget`.

CharWS:
https://github.com/simulationcraft/simc/wiki/ActionLists That'd be an example of syntax and the types of things that would want to be able to control in an action list (auto attack, snapshots, spells, pet controls, external buffs, consumables, sequences, waits, resources, conditional expressions, time, cooldown alignment/sync, target health, channeling, off gcd abilities)
https://github.com/simulationcraft/simc/pull/7158 And here would be an example of what that could look like in practice, if you follow one of the links in that thread and open up the Raw Input can see how the APL has been laid out

lime-green:
@jimmyt857 this APL feature is something I'm really excited about. It's going to be a huge undertaking so I think it has to be approached with incremental additions to the sim until we can replace rotations with an APL rotation. At a high level:

1. Add ability to optionally use an APL rotation engine (server flag)
2. Add default APL file to 1 or more specs starting with working syntax for extremely basic needs (auto attack, cast)
3. Keep adding to APL file and implementing APL engine logic to fulfill rotation needs until we can more or less match current rotation DPS
4. Add experimental flag in UI to enable APL rotation free-form text field 
5. Ask all spec developers to implement current rotation as APL file, fulfill APL engine logic as needed
6. Make APL rotation the default engine 
7. Remove old rotation engine 

I think it makes sense to follow the same APL syntax as simc unless there's a good reason not to, it's tried and true and will serve as a good yardstick. 

What do you think about this? I think #4 is quite important since it will empower users to create their own APLs and I think we'll be surprised at how eager people will be in min-maxing rotations. I believe once users get access to a basic APL for their spec a lot of the work can be done by them (having APL competitions in class discords?).

jimmyt857:
@lime-green the steps you list are all great, I'd just like to note 2 points:

[...]

PR #3180, opened 2023-06-19
First big chunk of UI work for APL

jimmyt857:
This sets up framework for actions UI. Biggest remaining piece of work is Value selection.


jimmyt857:
#2173

PR #3194, opened 2023-06-21
Fix the multiplicative stacking of armor buffs

merijn:
Implement a new interface for scaling equipment stats and update all armor scaling talents/metagems/forms/presence to use this new interface.


merijn:
After manually checking the DTPS and armor changes for all 4 classes, this LGTM.

PR #3340, opened 2023-07-18
crit debuffs should not snapshot

lime-green:


lime-green:
Not sure why I thought they didn't snapshot but Ketesh reminded me they do (and I now remember warlocks use that for corruption)

PR #3433, opened 2023-08-08
rogue:fix the tanlent "mace specialization"

joexi:
rogue:fix the tanlent "mace specialization"


joexi:
@vigo2 Is this okay?I've no idea how to edit last pr, so i make a new one

PR #3434, opened 2023-08-08
rogue:fix the tanlent "mace specialization"

joexi:
rogue:fix the tanlent "mace specialization"


joexi:
@vigo2 you are right , use "add stats" will be the best solution

vigo2:
LGTM. Merged.

PR #3465, opened 2023-08-16
fix APL wait

lime-green:
Fixes issue with current APL wait where it was reducing GCD timer eg:
image
(see plague strike + oblit being casted at the same time)

Now, with the fix:

With config:
image

image

Making wait time 2000ms:
image

### DIFF
--- a/sim/core/apl.go
+++ b/sim/core/apl.go
@@ -181,10 +181,8 @@ func (apl *APLRotation) DoNextAction(sim *Simulation) {
 }
 
 func (apl *APLRotation) getNextAction(sim *Simulation) *APLAction {
-	for _, unit := range sim.AllUnits {
-		if sim.CurrentTime < unit.waitUntilTime {
-			return nil
-		}
+	if sim.CurrentTime < apl.unit.waitUntilTime {
+		return nil
 	}
 
 	if apl.strictSequence != nil {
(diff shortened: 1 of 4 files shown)


sim/core/apl.go
```diff
@@ -181,6 +181,12 @@ func (apl *APLRotation) DoNextAction(sim *Simulation) {
 }
 
 func (apl *APLRotation) getNextAction(sim *Simulation) *APLAction {
+	for _, unit := range sim.AllUnits {
```
jimmyt857, line 184: Looping through all units on the hot path will be a noticeable performance loss. Seems like we can just check `apl.unit` here anyway right?

sim/core/unit.go
```diff
@@ -137,6 +137,7 @@ type Unit struct {
 	waitingForEnergy float64
 	waitingForMana   float64
 	waitStartTime    time.Duration
+	waitUntilTime    time.Duration
```
jimmyt857, line 140: This value needs to be put back to 0 on reset

sim/core/apl.go
lime-green: yep, thanks!

jimmyt857: I still see a loop over all units here, is there a missing commit?

jimmyt857 (changes requested):

jimmyt857 (approved):

PR #3576, opened 2023-08-30
Vigo/various cleanups

vigo2:
[core] add an "aggregator" type for calculating mean and standard deviation, used in various metrics

[core] add shortcut functions to access commonly used equips (e.g. MainHand(), OffHand() etc.)
[core] replace core.ItemSlot usages by proto.ItemSlot
[various] use said shortcut functions, and replace Get{MH,OH,Ranged}Weapon() where appropriate (e.g. WeaponType or Enchant checks)

[core] add GetProcMaskForTypes() to help the various weapon specializations
[warrior] refactor applyWeaponSpecializations() similar to rogue's

[tools] various simplifications for WowDatabase and io_utils

plus minor linting issues etc. all over


sim/deathknight/talents_blood.go
```diff
@@ -33,7 +33,7 @@ func (dk *Deathknight) ApplyBloodTalents() {
 	dk.applyScentOfBlood()
 
 	// Two Handed Specialization
-	if dk.HasMHWeapon() && dk.Equip[proto.ItemSlot_ItemSlotMainHand].HandType == proto.HandType_HandTypeTwoHand {
+	if dk.MainHand().HandType == proto.HandType_HandTypeTwoHand {
```
jimmyt857, line 36: I'm pretty sure calls like this will panic (due to `MainHand()` returning `nil`) when there is no item equipped for the slot

vigo2: These just return a pointer to an Equipment slot ([17]Item), so they're never nil.

jimmyt857 (approved):

PR #3620, opened 2023-09-05
vigo/advanced-advance

vigo2:
[core] rewrite execute phase handling to minimize the number of checks in advance()

[core] cache players and pets as *Character for use in advance()

[core] cleanup - cached pets are used instead of (renamed) PetAgents wherever applicable

[core] cleanup - change "isExecute" parameter from int to int32, for good measure

[core] cleanup - remove superfluous "not empty" check before iterating character.Pets in various places

[core] cleanup - remove unused parameters in init(), reset(), and advance() hierarchies

### DIFF
--- a/sim/core/character.go
+++ b/sim/core/character.go
@@ -82,6 +82,8 @@ type Character struct {
 	defensiveTrinketCD *Timer
 	offensiveTrinketCD *Timer
 	conjuredCD         *Timer
+
+	Pets []*Pet // cached in AddPet, for advance()
 }
 
 func NewCharacter(party *Party, partyIndex int, player *proto.Player) Character {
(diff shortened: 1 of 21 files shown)


sim/core/character.go
```diff
@@ -338,45 +340,31 @@ func (character *Character) AddPet(pet PetAgent) {
...
-	}
-	panic(character.Name + " has no pet with name " + name)
+	character.PetAgents = append(character.PetAgents, pet)
+	character.Pets = append(character.Pets, pet.GetPet())
 }
 
 func (character *Character) MultiplyMeleeSpeed(sim *Simulation, amount float64) {
 	character.Unit.MultiplyMeleeSpeed(sim, amount)
 
-	if len(character.Pets) > 0 {
```
jimmyt857, line 356: Have you checked perf changes on this, or at least an overall perf comparison for this PR?

vigo2: The len() checks before iterating have no influence on performance, at least with go 1.21. Overall, this is a 3-4% perf increase for dual-wielding classes (go 1.21, ARM64)

vigo2: Just retested for BM hunter: that's an even larger increase, 6-7%. Virtual calls are remarkably "expensive", and most of them are gone now for pets.

jimmyt857 (approved):

Issue #3653, opened 2023-09-10
Fetish of Volatile Power


CamSimons:
**Describe the bug**
On the sim for Enhancement Shaman, Fetish of Volatile Power is behaving incorrectly. The Volatility buff should stack on hits with Flametongue Weapon, not just with spellcasts.

**To Reproduce**
Equip Fetish of Volatile Power, enchant weapons with Flametongue. Activate the trinket and attack something.

**Expected behavior**
Auto attacking without casting spells should still stack the Volatility buff due to the Flametongue weapon procs. 

**Sim Links and Screenshots**
https://classic.warcraftlogs.com/reports/vQpHL12J9Y87jXDN#fight=last&view=events

PR #3660, opened 2023-09-11
flametongue can proc fetish

lime-green:
Closes https://github.com/wowsims/wotlk/issues/3653


Basskip:
@lime-green Not super familiar with the codebase, but looks like this change might also need to be separately applied to the alliance versions of the item?

https://www.wowhead.com/wotlk/item=47726/talisman-of-volatile-power
https://www.wowhead.com/wotlk/item=47946/talisman-of-volatile-power

PR #3702, opened 2023-09-19
Implement APL Channel action

jimmyt857:
channel


jimmyt857:
Test results changed due to mana ticks now having no effect during channels

PR #3709, opened 2023-09-20
vigo/dk-stuff

vigo2:
[dk] slight RuneCost and RunicPowerBar overhaul

There's a number of TODs added where the code seems suspicous or worse ;) These are for later.

@jimmyt857 Some "bindings" for APL are wrong, too - namely, APLValueCurrentRuneCount and APLValueCurrentNonDeathRuneCount return the same underlying value. Similarly, it's rather unclear to me what APLValueNextRuneCooldown should return - the currently used NextRuneTypeReadyAt() isn't actually doing what it's commenting.

### DIFF
--- a/sim/core/apl_values_runes.go
+++ b/sim/core/apl_values_runes.go
@@ -30,11 +30,11 @@ func (value *APLValueCurrentRuneCount) Type() proto.APLValueType {
 func (value *APLValueCurrentRuneCount) GetInt(_ *Simulation) int32 {
 	switch value.runeType {
 	case proto.APLValueRuneType_RuneBlood:
-		return int32(value.unit.CurrentBloodRunes()) // TODO these only count non-death runes of the specified kind
+		return int32(value.unit.CurrentBloodOrDeathRunes())
 	case proto.APLValueRuneType_RuneFrost:
-		return int32(value.unit.CurrentFrostRunes())
+		return int32(value.unit.CurrentFrostOrDeathRunes())
 	case proto.APLValueRuneType_RuneUnholy:
-		return int32(value.unit.CurrentUnholyRunes())
+		return int32(value.unit.CurrentUnholyOrDeathRunes())
 	case proto.APLValueRuneType_RuneDeath:
 		return int32(value.unit.CurrentDeathRunes())
 	}
--- a/sim/core/apl_values_runes.go
+++ b/sim/core/apl_values_runes.go
@@ -27,10 +27,10 @@ func (rot *APLRotation) newValueCurrentRuneCount(config *proto.APLValueCurrentRu
 func (value *APLValueCurrentRuneCount) Type() proto.APLValueType {
 	return proto.APLValueType_ValueTypeInt
 }
-func (value *APLValueCurrentRuneCount) GetInt(sim *Simulation) int32 {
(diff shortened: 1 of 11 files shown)


vigo2:
> It seems you have changed it to in fact call the same method in here tho and I dont understand why?

Because they had different names, but were - in fact - doing the same (returning the number of non-death, non-spent runes of the given kind). Are the number of non-spent runes (converted to death or not) of a given kind even useful?

> If you think its not doing what its commenting to do please do tell.

```
func (rp *RunicPowerBar) NextRuneTypeReadyAt(sim *Simulation, left int8, right int8) time.Duration {
	if rp.runeStates&isSpents[left] != isSpents[left] && rp.runeStates&isSpents[right] != isSpents[right] {
		// Both are ready so return current time
		return sim.CurrentTime
	} else if rp.runeStates&isSpents[left] == isSpents[left] || rp.runeStates&isSpents[right] == isSpents[right] {
		// One is spent so return the time it will regen at
		return MinDuration(rp.runeMeta[left].regenAt, rp.runeMeta[right].regenAt)
	}
	// Both are spent so return the last one to regen at
	return MaxDuration(rp.runeMeta[left].regenAt, rp.runeMeta[right].regenAt)
}
```
This will never reach the `// Both are spent...` part since the `// One is spent...` part is actually `// At least one is spent...`, so it works. "BothXXXRunesReadyAt()" is a pretty clear name (albeit a bit long, though "Rune" or "Runes" is redudant in this context, anyhow).

rosenrusinov:
> Because they had different names, but were - in fact - doing the same (returning the number of non-death, non-spent runes of the given kind). Are the number of non-spent runes (converted to death or not) of a given kind even useful?

If they were doing the same then that is a mistake in the internals of the called method. I think instead of changing what the apl valuea are supposed to do we should fix them to do it instead. About the usability pf the values i think i already used them in some apls and having that extra information wont hurt

> This will never reach the // Both are spent... part since the // One is spent... part is actually // At least one is spent..., so it works. "BothXXXRunesReadyAt()" is a pretty clear name (albeit a bit long, though "Rune" or "Runes" is redudant in this context, anyhow).

I see what did wrong and in fact it has to check if its only one active rather then atleast one and should be changed to that.

jimmyt857 (approved):
I have no concerns since there are no test breakages here. As long as @rosenrusinov's concerns are resolved, LGTM.

rosenrusinov:
> I have no concerns since there are no test breakages here. As long as @rosenrusinov's concerns are resolved, LGTM.

There are no tests on the DK that are using the new APL which is why you see no test brekages ^^ That reminds me I should set up some tests with the APLs now that we have them

vigo2:
I'm going to fix both pre-existing bugs (i.e. `NextRuneTypeReadyAt()`, and add a set of `CurrentXXXOrDeathRunes()` functions for APL). Since `CurrentXXXRunes()` was used almost everywhere (possibly assuming it included death runes or not), I'm leaving the existing rotation code "as is".

> About the usability of the values i think having that extra information wont hurt

I think it does, as we've basically just seen: there's so many subtle variations for rune counts or rune ready times that clients have a hard time choosing the (possibly) right one.

PR #3714, opened 2023-09-20
vigo/cast

vigo2:
[core] makeCastFunc() outlined, but specialized for some common setups, plus some cleanups

[dk] gorgoyle now recasts in ApplyEffects() directly, removing the last Cast.OnCastComplete() use case

Contrary to intuition, this is actually a perf gain.

Question for a followup-PR: Is it ok to remove `PseudoStats.GracefulCastCDFailures`, or will this cause too many issues (crashes) with the legacy DK rotations? Just sprinkling two `Cast.CD.IsReady()` allows all DK tests to pass.

### DIFF
--- a/sim/core/spell.go
+++ b/sim/core/spell.go
@@ -168,14 +168,29 @@ func (unit *Unit) RegisterSpell(config SpellConfig) *Spell {
 	// Default the other damage multiplier to 1 if only one or the other is set.
 	if config.DamageMultiplier != 0 && config.DamageMultiplierAdditive == 0 {
 		config.DamageMultiplierAdditive = 1
-	}
-	if config.DamageMultiplierAdditive != 0 && config.DamageMultiplier == 0 {
+	} else if config.DamageMultiplierAdditive != 0 && config.DamageMultiplier == 0 {
 		config.DamageMultiplier = 1
 	}
 
 	if unit.IsUsingAPL {
 		config.Cast.DefaultCast.ChannelTime = 0
 		config.Cast.DefaultCast.AfterCastDelay = 0
+	}
+
+	if (config.DamageMultiplier != 0 || config.ThreatMultiplier != 0) && config.ProcMask == ProcMaskUnknown {
+		panic("ProcMask for spell " + config.ActionID.String() + " not set")
+	}
+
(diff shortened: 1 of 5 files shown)


sim/core/spell.go
```diff
@@ -178,6 +177,22 @@ func (unit *Unit) RegisterSpell(config SpellConfig) *Spell {
 		config.Cast.DefaultCast.AfterCastDelay = 0
 	}
 
+	if (config.DamageMultiplier != 0 || config.ThreatMultiplier != 0) && config.ProcMask == ProcMaskUnknown {
+		panic("ProcMask for spell " + config.ActionID.String() + " not set")
+	}
+
+	if (config.DamageMultiplier != 0 || config.ThreatMultiplier != 0) && config.SpellSchool == SpellSchoolNone {
+		panic("SpellSchool for spell " + config.ActionID.String() + " not set")
```
jimmyt857, line 185: I feel like there are spells somewhere with no school. I know gift of the naaru (draenei racial) has no school, for example, although we haven't implemented it yet.

sim/core/spell.go
```diff
@@ -450,10 +476,10 @@ func (spell *Spell) CanCast(sim *Simulation, target *Unit) bool {
 		return false
 	}
 
-	// While casting, no other action is possible
+	// While casting or channeling, no other action is possible
```
jimmyt857, line 479: This is no longer the correct condition to check for channels, so it's intentional for the comment to not mention channels.

Also, channeling doesn't prevent casts, that's what recasts are. So it's important that this function not return false just because the agent is channeling. I ran into this recently while adding the Channel APL action.

jimmyt857 (approved):

sim/core/spell.go
vigo2: Channeling still sets a "hardcast" though, so effectively it _does_ prevent other cast actions.

By "recast" I assume you mean cancelling the current "hardcast", then casting again? That's fine, but doesn't affect this comment, because then there's no "hardcast" happening anymore.

sim/core/spell.go
vigo2: This check is pre-existing, so I'd leave it "as is". Adding more static checks here helps moving stuff out of the critical path, so this is a good thing.

PR #3718, opened 2023-09-21
updated a bunch of weird little things with lust and apl

lologarithm:


sim/core/buffs.go
```diff
@@ -596,6 +596,14 @@ func registerBloodlustCD(agent Agent) {
 
 func BloodlustAura(character *Character, actionTag int32) *Aura {
 	actionID := BloodlustActionID.WithTag(actionTag)
+
+	sated := character.GetOrRegisterAura(Aura{
```
lologarithm, line 600: sated was added because we want to be able to control a shaman casting bloodlust in the APL.  If you have the raid buff also selected and have both cast in the APL, we should prevent the second one from working.

sim/core/mana.go
```diff
@@ -262,7 +262,8 @@ func (sim *Simulation) initManaTickAction() {
 			char := player.GetCharacter()
 			char.ManaTick(sim)
 			if char.OnManaTick != nil {
-				if char.IsUsingAPL {
+				// Only execute APL actions after mana ticks once pre-pull has completed.
+				if char.IsUsingAPL && sim.CurrentTime > 0 {
```
lologarithm, line 266: prevent using main rotation during pre-pull

sim/shaman/bloodlust.go
```diff
@@ -12,7 +12,7 @@ func (shaman *Shaman) BloodlustActionID() core.ActionID {
 }
 
 func (shaman *Shaman) registerBloodlustCD() {
-	if !shaman.SelfBuffs.Bloodlust {
+	if !shaman.SelfBuffs.Bloodlust && !shaman.IsUsingAPL {
```
lologarithm, line 15: we want the shaman lust spell to be always registered as an option if using the APL (even if the settings UI has disabled bloodlust)

sim/shaman/bloodlust.go
```diff
@@ -60,9 +61,11 @@ func (shaman *Shaman) registerBloodlustCD() {
 		},
 	})
 
-	shaman.AddMajorCooldown(core.MajorCooldown{
-		Spell:    bloodlustSpell,
-		Priority: core.CooldownPriorityBloodlust,
-		Type:     core.CooldownTypeDPS,
-	})
+	if !shaman.IsUsingAPL {
+		shaman.AddMajorCooldown(core.MajorCooldown{
```
lologarithm, line 65: if using the APL, we want to actually cast the spell, not have it auto used with other MCDs

ui/balance_druid/presets.ts
```diff
@@ -278,7 +278,7 @@ name: 'Basic P3 APL',
 rotation: SavedRotation.create({
 	specRotationOptionsJson: BalanceDruidRotation.toJsonString(DefaultRotation),
 	rotation: APLRotation.fromJsonString(`{
-      "type": "TypeAPL",
+      "type": 3,
```
lologarithm, line 281: turns out the string there was not actually parsing correctly, causing the 'IsUsingAPL' flag to not be set correctly.

ui/core/proto_utils/action_id.ts
```diff
@@ -339,11 +339,13 @@ export class ActionId {
 			case 'Mana Tide Totem':
 			case 'Power Infusion':
 				if (this.tag != -1) {
-					if (this.tag === playerIndex) {
+					if (this.tag === playerIndex || playerIndex == undefined) {
```
lologarithm, line 342: individual sim playerIndex was undefined, this is simple check to show (self) on self cast

PR #3779, opened 2023-09-29
updated to use go 1.21

lologarithm:


sim/core/bulksim_test.go
```diff
@@ -107,7 +107,15 @@ func TestBulkSim(t *testing.T) {
...
-		return goproto.MarshalTextString(a) == goproto.MarshalTextString(b)
+		am, err := goproto.Marshal(a)
+		if err != nil {
+			return false
+		}
+		bm, err := goproto.Marshal(b)
+		if err != nil {
+			return false
+		}
+		return string(am) == string(bm)
```
vigo2, line 118: protojson.Format() is probably a better fit

go.mod
```diff
@@ -1,16 +1,15 @@
 module github.com/wowsims/wotlk
 
-go 1.18
+go 1.21
 
 require (
-	github.com/golang/protobuf v1.5.3
 	github.com/google/go-cmp v0.5.8
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
 	github.com/satori/go.uuid v1.2.0
```
vigo2, line 8: should be replaced by github.com/google/uuid v1.3.1. the single use-case is in web/main.go:139, and becomes `newID := uuid.NewString()`.

go.mod
```diff
@@ -1,16 +1,15 @@
...
 	github.com/spf13/cobra v1.7.0
 	github.com/tailscale/hujson v0.0.0-20221223112325-20486734a56a
 	golang.org/x/exp v0.0.0-20221028150844-83b7d23a625f
```
vigo2, line 11: slices are now part of the standard library (so it's just "slices" for importing). if you update this, too, note that you'll have to update `slices.SortFunc()` etc., since they've changed - sadly - from using Less() to Cmp(): e.g.

```
	slices.SortFunc(items, func(a, b Item) bool {
		return a.ID < b.ID
	})
```
becomes
```
	slices.SortFunc(items, func(a, b Item) int {
		return cmp.Compare(a.ID, b.ID)
	})
```

vigo2 (approved):

PR #3787, opened 2023-09-30
Make suggest gemming shared & add for hunter

1337LutZ:
Tried to make the suggest option more generic to be more easily implemented by other classes.
(I know the feature in general was/is pretty tailored towards ArP users that rely on quite "flat" caps)

- Moved all auto gem code to shared folder
- Incorporated existing spec (Feral/Warrior) specific logic
- Added Hunter (need a check on this but it seems to work pretty well vs existing BIS sets) 
- Tweaked Warrior gemming to allow for gemming ArP in yellow sockets


hululula:
Worked with 1337LutZ to verify the gemming configurations and it looks good to me now.
- Monsterik in discord

PR #3789, opened 2023-09-30
Jake/apl tests

jarveson:
- rip apl presets into 'json' files, so that they can be pulled into test functions and tested


lologarithm:
idea to go with a PR I am cooking up, name all apl files 

XXX.apl.json -- this will let you edit those files with auto-completion once I get this readme in on how to make autocomplete work!

sim/deathknight/dps/dps_deathknight_test.go
```diff
@@ -12,6 +14,16 @@ func init() {
 	RegisterDpsDeathknight()
 }
 
+func GetAplRotation(dir string, file string) core.RotationCombo {
```
jimmyt857, line 17: It looks like this same function is in every test file, can it be placed in one of the testing files in `core` and shared?

jarveson: good idea, moved it into test_utils

jarveson:
@lologarithm renamed them as suggested

PR #3802, opened 2023-10-02
more-min-max

vigo2:
[core] change MajorCooldown.Priority from float64 to int32
[everywhere] replaced core.MinDuration, core.MaxDuration, and core.MaxTristate by min/max as well [everywhere] changed some min(min(...)) and similar constructs to use multi-value min(...) instead


sim/core/major_cooldown.go
```diff
@@ -507,15 +507,10 @@ func (mcdm *majorCooldownManager) UpdateMajorCooldowns() {
...
-		if m1r != m2r {
-			return int(m1r - m2r)
+		if r1, r2 := m1.ReadyAt(), m2.ReadyAt(); r1 != r2 {
+			return int(r1 - r2)
 		}
-		if m1.Priority > m2.Priority {
-			return -1
-		}
-		return 1
+		return int(m2.Priority - m1.Priority)
```
lologarithm, line 513: so this does actually change tests... dunno if its worth caring about

sim/shaman/totems.go
```diff
@@ -188,11 +188,7 @@ func (shaman *Shaman) registerCallOfTheElements() {
 }
 
 func (shaman *Shaman) NextTotemAt(_ *core.Simulation) time.Duration {
-	nextTotemAt := core.MinDuration(
-		core.MinDuration(shaman.NextTotemDrops[0], shaman.NextTotemDrops[1]),
-		core.MinDuration(shaman.NextTotemDrops[2], shaman.NextTotemDrops[3]))
-
-	return nextTotemAt
+	return min(shaman.NextTotemDrops[0], shaman.NextTotemDrops[1], shaman.NextTotemDrops[2], shaman.NextTotemDrops[3])
```
lologarithm, line 191: nice

sim/core/major_cooldown.go
vigo2: hmm, it didn't for me, but most likely could, since it now returns 0 for equal priorities (which is "more correct", in a way).
none of these "x1 - x2" things are "strictly correct", since they fail when "x1 - x2" overflows (e.g. MaxInt32 - (-MaxInt32) or stuff like that), but it won't really matter. currently, using cmp.Compare(x1, x2) is (microbenchmarky, but still) slower, but handles even the NaN-cases in a predictable way.

lologarithm: Right, which is what I was trying to summarize haha

lologarithm (approved):

sim/core/major_cooldown.go
lologarithm: And you do have test changes in this pr

PR #3837, opened 2023-10-07
Simple-perf-measures

vigo2:
[core] don't needlessly reset auraTracker.minExpires fully after Refresh() or Deactivate() anymore

[character] advance() assumes that pets don't have pets anymore

I'm planning to make some more changes building on the latter assumption - does that sound viable?


jimmyt857 (approved):
Yes AFAIK it's a valid assumption that pets won't have pets.

PR #3847, opened 2023-10-08
Add melee crit cap display

### LINKED ISSUE #2774: Crit Cap for Auto Attack Display
**Is your feature request related to a problem? Please describe.**
For dual wielding auto attacks, your crit chance is capped lower than your abilities crit chance - autos are single roll, which can miss or glance, which cannot crit. This means that each specific simulation character with different hit/exp/crit has different auto attack crit caps, and will run into strange simulation results when going over or under their crit cap, if they are unaware of the mechanic, or simply if they didn't realize a specific setting/gear gave them wasted crit past their cap, or increased their cap and realized more crit from simply adding some hit.

As an example, the regular phase 2 BIS ring for an Assassination Rogue is Loop of the Agile, which gives crit rating, and is slightly better in almost all setups over Band of Lights, which gives hit rating. If you specifically have fairly low hit and the Dark Matter trinket (which gives a large amount of crit on proc), then you are sufficiently over crit cap during the Dark Matter proc and Band of Lights will suddenly sim better than Loop of the Agile. 

**Describe the solution you'd like**
I am not sure the best implementation, but for my data-hungry brain I would love to see a breakdown of:
1. Auto attack crit cap statically
2. Auto attack crit cap during crit-increasing temporary buffs available to the current simulation character
3. Your crit in relation to both of those
Maybe when hovering over your "Melee Crit" stat on the display tab.

That being said, knowing which procs give you crit is likely too much hardcoding, and maybe the sim should just display it in the results tab somehow - maybe measure each auto attacks attack table, and if you're crit capped, note a number of auto attacks that were X% over crit cap? Some sort of warning when the sim runs and finds itself over crit cap?

**Additional context**
I mostly write this to raise awareness and start brainstorming, this is an issue that is only very rarely relevant now, but will become more relevant as the public starts to sim higher crit chance characters eg. Icecrown Citadel gearing.

secretbis:
Added in the form of a new spec-gated row in the character stats pane

Text is green if below crit cap, red if above, white if equal.  

Tooltip includes several details about said crit cap, including mechanics modifiers, cap final value, and how much the cap can be raised.

A special modifier for Enhancement is included, based on the number of talent points in Elemental Devastation.  The stats display now updates on talent change to accommodate this change.  

Screenshots:

image

image

image

Resolves #2774

### DIFF
--- a/ui/core/components/character_stats.tsx
+++ b/ui/core/components/character_stats.tsx
@@ -1,4 +1,4 @@
-import { Stat, Class, PseudoStat } from '..//proto/common.js';
+import { Stat, Class, PseudoStat, Spec } from '..//proto/common.js';
 import { TristateEffect } from '..//proto/common.js'
 import { getClassStatName, statOrder } from '..//proto_utils/names.js';
 import { Stats } from '..//proto_utils/stats.js';
@@ -19,6 +19,7 @@ export type StatMods = { talents: Stats };
 export class CharacterStats extends Component {
 	readonly stats: Array<Stat>;
 	readonly valueElems: Array<HTMLTableCellElement>;
+	readonly meleeCritCapValueElem: HTMLTableCellElement | undefined;
 
 	private readonly player: Player<any>;
 	private readonly modifyDisplayStats?: (player: Player<any>) => StatMods;
@@ -45,7 +46,7 @@ export class CharacterStats extends Component {
 			const row = (
 			<tr
 				className='character-stats-table-row'
-			>	
+			>
 				<td className="character-stats-table-label">{statName}</td>
 				<td className="character-stats-table-value">
 					{this.bonusStatsLink(stat)}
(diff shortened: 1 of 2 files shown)


ui/core/components/character_stats.tsx
```diff
@@ -262,4 +348,26 @@ export class CharacterStats extends Component {
 
 		return link as HTMLElement;
 	}
+
+	private shouldShowMeleeCritCap(player: Player<any>): boolean {
+		return [
```
secretbis, line 353: The row will only display for specs where a melee crit cap is relevant

ui/core/components/character_stats.tsx
```diff
@@ -45,7 +46,7 @@ export class CharacterStats extends Component {
 			const row = (
 			<tr
 				className='character-stats-table-row'
-			>	
+			>
```
secretbis, line 49: My editor auto-trims trailing whitespace - there's a few diff lines of that here

ui/core/components/character_stats.tsx
```diff
@@ -58,8 +59,23 @@ export class CharacterStats extends Component {
...
+
+			table.appendChild(row);
+			this.meleeCritCapValueElem = row.getElementsByClassName('character-stats-table-value')[0] as HTMLTableCellElement;
+		} else {
+			this.meleeCritCapValueElem = undefined;
+		}
+
 		this.updateStats(player);
-		TypedEvent.onAny([player.currentStatsEmitter, player.sim.changeEmitter]).on(() => {
+		TypedEvent.onAny([player.currentStatsEmitter, player.sim.changeEmitter, player.talentsChangeEmitter]).on(() => {
```
secretbis, line 78: Character stats now update on talent change to compute Elemental Devastation's effect on crit cap

ui/core/player.ts
```diff
@@ -657,6 +674,61 @@ export class Player<SpecType extends Spec> {
...
+			// Elemental Devastation uptime is near 100%
+			const ranks = (this as Player<Spec.SpecEnhancementShaman>).getTalents().elementalDevastation;
+			specSpecificOffset = 3.0 * ranks;
+		}
+
+		let debuffCrit = 0.0;
+
+		const debuffs = this.sim.raid.getDebuffs();
+		if (debuffs.totemOfWrath || debuffs.heartOfTheCrusader || debuffs.masterPoisoner) {
+			debuffCrit = 3.0;
```
secretbis, line 704: Crit from debuffs has to be manually adjusted here -the display seems to handle adjusting that, and not the player abstraction.  

I've only spent about two hours with this codebase, so let me know if there's a better way to find that total aggregate value.

lime-green:
Great job; I think there could be a few follow-up improvements: white vs special hit cap, remove DKs (don't think crit cap is relevant), add boomies (and possibly more ranged classes) but let's see how people respond with this out there!

PR #3851, opened 2023-10-08
fix weird tricks + MCD interaction

lime-green:
Was causing rogue cooldowns not to be cast if tricks was cast pre-pull

### DIFF
--- a/sim/core/major_cooldown.go
+++ b/sim/core/major_cooldown.go
@@ -183,8 +183,6 @@ type majorCooldownManager struct {
 	// the course of the sim.
 	majorCooldowns []*MajorCooldown
 	minReady       time.Duration
-
-	tryUsing bool
 }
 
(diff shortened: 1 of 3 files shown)


sim/core/major_cooldown.go
```diff
@@ -486,17 +483,12 @@ restart:
...
 	}
-	mcdm.tryUsing = false
-
 	mcdm.minReady = mcdm.majorCooldowns[0].ReadyAt()
 }
 
 // This function should be called if the CD for a major cooldown changes outside
 // of the TryActivate() call.
 func (mcdm *majorCooldownManager) UpdateMajorCooldowns() {
-	if mcdm.tryUsing {
```
lime-green, line 497: afaik it shouldn't be a big deal if this happens, and occurs for tricks because tricks needs to update MCD on its cast

PR #3859, opened 2023-10-09
ghoul-updates

vigo2:
[dk] ghoul's Claw ability now has a 1s GCD, isn't affected by haste anymore, and has a refund, as tested in-game (it's essentially a Sinister Strike)

I can't test at the moment, but it looks like non-permanent ghouls use the pre-WotLK energy ticks (~20e / 2s), while the permanent ghoul gains energy every ~100ms. None of that is very consequential, of course ;)

### DIFF
--- a/sim/deathknight/ghoul_pet_abilities.go
+++ b/sim/deathknight/ghoul_pet_abilities.go
@@ -2,6 +2,7 @@ package deathknight
 
 import (
 	"github.com/wowsims/wotlk/sim/core"
+	"time"
 )
 
 type PetAbilityType int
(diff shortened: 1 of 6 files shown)


sim/deathknight/ghoul_pet_abilities.go
```diff
@@ -61,8 +62,9 @@ func (ghoulPet *GhoulPet) newClaw() PetAbility {
 
 			Cast: core.CastConfig{
 				DefaultCast: core.Cast{
-					GCD: core.GCDDefault,
+					GCD: time.Second,
```
rosenrusinov, line 65: isn't it better to use `core.GCDMin` which is 1 second

rosenrusinov (approved):

sim/deathknight/ghoul_pet_abilities.go
vigo2: I've gone with the rogue house-style, that's simply using `time.Second`. `core.GCDMin` is essentially a "casts affected by cast speed limit", and doesn't quite apply here.

PR #3884, opened 2023-10-15
vigo/resource-tasks

vigo2:
[core] focus and energy ticks are not using PAs, but an advance()-like system

[dk] ghouls now use the regular focus bar, and regular spell casting


sim/core/focus.go
```diff
@@ -60,16 +61,15 @@ func (fb *focusBar) AddFocus(sim *Simulation, amount float64, actionID ActionID)
 	}
 }
 
-func (fb *focusBar) SpendFocus(sim *Simulation, amount float64, metrics *ResourceMetrics) {
+func (fb *focusBar) SpendFocus(sim *Simulation, amount float64, actionID ActionID) {
```
jimmyt857, line 64: Why are we switching to ActionID here? Metrics as input forces the creation of a proper Metrics object, which ensures all resource changes are properly exposed in the UI (timeline, resource charts, etc)

vigo2: The removed `ghoul_focus` didn't support metrics at all (apart from the fact that ghouls use energy to start with...), and for hunters, the metrics were used for spending, but not for gaining focus.
Since I'm leaning on the perf side at the moment ;) (and there's no pet rotation logic to debug via UI), I've decided to remove it entirely.

vigo2: I've taken a look at those fishy looking Subtlety tests where DPS decreased by >10%: some iterations stopped short because they had no pending actions anymore, only energy tick "tasks". I've added a sentinel pending action to protect against this situation - if auto attacks are modelled as "tasks", too, this might happen even in practice.

vigo2: Do you think it's viable to convert the current pre-pull logic to the following scheme: Add each prepull action as a PA with high priority (`ActionPriorityPrepull = 5` plus offsets to keep their configured order), then add a "StartPull" PA for time 0 (at `ActionPriorityPrepull`), and finally start a regular sim run at `PrepullStartTime()`?

jimmyt857: I'm not quite following why the issue is occurring, and how the proposed scheme addresses it. When the prepull runs out of actions, it should advance to `t = 0` and start the encounter. That advancement should be able to apply any energy gains. Maybe some logic is missing here? https://github.com/wowsims/wotlk/blob/master/sim/core/sim.go#L381

vigo2: The proposed pre-pull scheme is not for fixing any bugs, just for simplifying pre-pull handling (which is currently adding a second use-case to `runPendingActions()`, `Step(), and `advance()`).

The addressed issue was that some Subtlety tests were somehow running out of non-tick PAs. This was fixed by adding the sentinel PA (with `NextActionAt = NeverExpires`), so that that `Step()` never early exits until the actual end-of-combat conditions apply.
This in turn broke some pre-pulls, since those require `Step()` runs to stop at `max` which would now run happily until end-of-combat if only the sentinel PA was queued. That's why the `RunPrepullActions()` wrapper for `RunPendingActions()` has been added.

PR #3966, opened 2023-10-29
starfall tick dont proc omen

lime-green:


sim/druid/druid.go
```diff
@@ -206,11 +205,6 @@ func (druid *Druid) Initialize() {
 	druid.registerRebirthSpell()
 	druid.registerInnervateCD()
 	druid.registerFakeGotw()
-
-	if druid.RaidBuffTargets == 0 {
```
lime-green, line 210: this is always 30 now

PR #3985, opened 2023-10-31
Vigo/auto attacks

vigo2:
[core] introduce WeaponAttack, a per weapon (MH, OH, Ranged) AutoAttack

[warrior] ranged attacks (namely Thunder Clap) now properly trigger MH Deep Wounds

No test changes but for Arms-Warrior.

[core] handle WeaponAttacks via a "task"-like system

[core] slightly speedup advance() by using an inlineable tryAdvance() stub

Affected "Average" tests are mostly unchanged. For e.g. DKs and Enhancement Shamans, this is almost a ~10% perf gain (non-APL).

### DIFF
--- a/sim/core/sim.go
+++ b/sim/core/sim.go
@@ -54,6 +54,9 @@ type Simulation struct {
 	minTrackerTime time.Duration
 	trackers       []*auraTracker
 
+	minWeaponAttackTime time.Duration
+	weaponAttacks       []*WeaponAttack
+
 	minTaskTime time.Duration
 	tasks       []Task
 }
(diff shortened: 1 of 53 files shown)


sim/core/sim.go
```diff
@@ -450,6 +472,14 @@ func (sim *Simulation) Step() bool {
 	last := len(sim.pendingActions) - 1
 	pa := sim.pendingActions[last]
 
+	if pa.NextActionAt >= sim.minWeaponAttackTime && sim.minWeaponAttackTime <= sim.minTaskTime {
```
jimmyt857, line 475: It seems like we now have this idea of 'minimum timestamp amongst a group of events' in a few places (at least here and in aura trackers). Perhaps there would be a way to write a generic utility to manage it? Would definitely be in a separate PR.

jimmyt857 (approved):

sim/core/sim.go
vigo2: The "Task" interface and related methods was basically a shot at such a generic utility, but it's not quite there yet:
- it doesn't allow grouping tasks, which is important since grouping allows to minimize the costs of finding a "global" minimum
- it doesn't allow to inline the "ready" checks
Both problems can be tackled, of course.

PR #4029, opened 2023-11-11
Add blood tap as a major defensive CD option when 4pc T10 is present

Cyzin1:
t10 4pc buff support exists but the sim in legacy mode doesn't support blood tap as a Cooldown to schedule. This change allows for that and repurposes the old "Defensive" option to do something.


rosenrusinov (approved):

Cyzin1:
added commit updating the tooltip for bt priority to say T10 instead of T9

PR #4067, opened 2023-11-27
Change SOD pathing to use /classic/sod instead of just /classic

kayla-glick:


CyberQuackGames (approved):
Lgtm, everything loaded / ran fine.

Issue #4062, opened 2023-11-25
v0.1.57 prot warrior sim Crash Report 197143411


macwille:
## Crash on Simulate prot warrior 
Link:
http://localhost:3333/wotlk/protection_warrior/#eJztU1tIFFEYnv+fddwZFE+L1El0m4YSSaTZGXwokDMFEb2ZEkUPUQ9JDwXWk9nLlom3oEA3ckO2K1REBEFsgSEptfZQwkpiEuQFtKJcL5tLUnRmZy3CpFJ762Mu/+37/49zUda7kaCOh6AS/QA3AHoA6hHGAGKACcAeRD/C1lUEVLSgEuqTRZcQ3thp2IwEPEJOh3QKpNLDB44fPOaWiZwzCkoWefEZ6bs8TbrWiUNhVGTSNIOaaJtZpC4g0jNeTYqEkT9KBnk0hTSQq4mhkJ1uPifSBKcOOelMcq9JpI1ezXXhAQoKIbMtIm3xaulOXlA8ZHwY6dlcTeb8FMlDgiNIP+ZpshPgVN45wevO52oS9xxRIw2iJkaS5vRkSp+LDEza36cTdrgtkKrIIM28Y4PX8Vzk9jRqWcRF0/TMUnGPWGUvzhWMo9KBwn8sCR79bxnVkhsoqnALJEXIQRUbYbduGrrJX8M0zWKfrnPL8Bm6zzRMn1HsKzKNYl3nrj4fRb+IzWEIqDsURdIWRdoaRbUmigWno6jXRnEMXDFID4LwGJFCBCMQBngF2fIuG6F2a4WjFi1VCqLQnzoo+ZbWJMpJ018xV5NtrQ622uhmm5xID1tbkcRLZgygWk7TNvzjo1bQYC1HG39312/6lC/LnMXAWdv2cWuL4yfYfihLXuLB6uw72wcTbN/DHRlYKFnD1y92qXdxyUpP2G2re9n9qS+9VW8/MWeXP7CNr9dJJwMDzJkbY0rf1coSZYYt1MdT0/Hs8rYn3/NHS2yGYq2pfb8z7I2zwrqV/c9n4oyL55bL+sqn5fvSFq0/Kaevc0E98/TpR0Znb06U0T+euHeBax9jP/9/4Bu3qejF

RNG Seed: 2357535875

runtime error: invalid memory address or nil pointer dereference
Stack Trace:
goroutine 175 [running]:
[...]

PR #4073, opened 2023-12-02
Improve panic message for invalid talents, and fix prot warrior bug w…

jimmyt857:
…ith Rend not being registered


jimmyt857:
Fixes #4062

PR #4083, opened 2023-12-06
Object-oriented refactor of Suggest Gems

NerdEgghead:


NerdEgghead:
@1337LutZ @TheGroxEmpire Finally finished the Suggest Gems refactor to make it object-oriented and extensible to different types of sims without copy-pasting code. Would appreciate if you can take a look before I merge.

ui/feral_druid/sim.ts
```diff
@@ -334,231 +247,4 @@ export class FeralDruidSimUI extends IndividualSimUI<Spec.SpecFeralDruid> {
...
-				gemId = 42143;
-			}
-
-			updatedGear = updatedGear.withGem(redSocketList[i][0], redSocketList[i][1], this.sim.db.lookupGem(gemId));
-		}
-
-		return updatedGear;
-	}
-
-	detectArpStackConfiguration(arpTarget: number): boolean {
```
1337LutZ, line 560: You have this in the hunter sim, but it was removed in the Feral one.

ui/core/components/suggest_gems_action.ts
```diff
@@ -0,0 +1,627 @@
...
+	statCaps: Stats
+}
+
+interface SocketData {
+	itemSlot: ItemSlot
+	socketIdx: number
+}
+
+abstract class GemOptimizer {
+	protected readonly player: Player<any>;
```
1337LutZ, line 21: Any or `Spec` ? Since it always needs to be a typeof Spec?

ui/core/components/suggest_gems_action.ts
```diff
@@ -0,0 +1,627 @@
...
+		40111: 42142,
+		40119: 36767,
+	};
+
+	constructor(simUI: IndividualSimUI<any>) {
+		this.player = simUI.player;
+		this.sim = simUI.sim;
+
+		// Initialize empty arrays of gem priorities for each socket color
+		this.gemPriorityByColor = {} as Record<GemColor, Array<GemCapsData>>;
```
1337LutZ, line 45: Cant this be a empty object by default on line 23?
```
protected readonly gemPriorityByColor: Record<GemColor, Array<GemCapsData>> = {};
```

ui/core/components/suggest_gems_action.ts
```diff
@@ -0,0 +1,627 @@
...
+		}
+
+		return optimalSlot;
+	}
+	
+	socketGemInFirstMatchingSocket(gear: Gear, itemSlot: ItemSlot | null, colorToMatch: GemColor, gemId: number): Gear {
+		if (itemSlot != null) {
+			const item = gear.getEquippedItem(itemSlot);
+
+			for (const [socketIdx, socketColor] of item!.allSocketColors().entries()) {
```
1337LutZ, line 151: instead of item! do we want do just return early if `!item` on line 149-150?

ui/core/components/suggest_gems_action.ts
```diff
@@ -0,0 +1,627 @@
...
+		
+		/*
+		 * For specs that gem ArP, determine whether the current gear
+		 * configuration will optimally hard stack Fractured gems or not.
+		 */
+		this.passiveArp = passiveStats.getStat(Stat.StatArmorPenetration);
+		this.arpStackDetected = this.detectArpStackConfiguration(ungemmedGear);
+
+		// Update red gem priority
+		const redGemCaps = new Array<GemCapsData>();
```
1337LutZ, line 325: The priority of adding the gems on line 328 and below can be different per class/spec. Is there a way to also make that dynamic?

1337LutZ:
Super nice work <3 Added some comments in regards of TS / gem orders.
Will build it locally and see if it behaves the same as the current one :)

1337LutZ:
Created a build and looks like it is gemming sometimes even better (esp prioing the tear socket) for most specs 👍

ui/core/components/suggest_gems_action.ts
NerdEgghead: I wanted to initialize all the keys explicitly with empty Arrays, since I query the length elsewhere, such as on line 82.

ui/core/components/suggest_gems_action.ts
NerdEgghead: Do you have an example of a physical DPS spec that would use a different priority than the one listed here? My thought was to generalize code like this on an as-needed basis, as more sims start to implement GemOptimizer subclasses.

ui/core/components/suggest_gems_action.ts
NerdEgghead: My initial assumption was that `item` would always be non-null if `itemSlot` is non-null, since `findStrongestSocketBonus()` already has a null check in it. But I added the early return anyway just for completeness.

PR #4095, opened 2023-12-11
Start implementation of classic world buffs

kayla-glick:
# Background
I added all major world buffs (list taken from 60U) besides Resist Fire to a new "World Buffs" section on the settings page. I did my best to implement the buffs but there were certain cases where I wasn't quite sure what the right way to approach the implementation was. I left TODOs for each of these. Toggling buffs works as expected and you can see the character's stats change as they're toggled.

image

image

# Open Issues
### Wowhead tooltips for Sayge's buffs don't display the correct stats tooltip
image

### Ranged Crit chance is missing as a stat


kayla-glick:
Made Sayge's an enum picker and edited the picker a bit so that you can get the icon while no buff is selected. Also changed how text is displayed for the picker to be more consistent with standard icon pickers.

image
image

PR #4099, opened 2023-12-15
Updated rend.go to allowing usage while Defensive Stance

m1ker1n:
Rend requires Battle Stance OR Defensive Stance.

Made PR to fix this because couldn't add Rend into rotation for prot warrior

### DIFF
--- a/sim/warrior/rend.go
+++ b/sim/warrior/rend.go
@@ -34,7 +34,7 @@ func (warrior *Warrior) RegisterRendSpell(rageThreshold float64, healthThreshold
 		},
 
 		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
-			return warrior.StanceMatches(BattleStance)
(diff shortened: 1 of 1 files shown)


sim/warrior/rend.go
```diff
@@ -34,7 +34,7 @@ func (warrior *Warrior) RegisterRendSpell(rageThreshold float64, healthThreshold
 		},
 
 		ExtraCastCondition: func(sim *core.Simulation, target *core.Unit) bool {
-			return warrior.StanceMatches(BattleStance)
+			return warrior.StanceMatches(BattleStance) || warrior.StanceMatches(DefensiveStance)
```
kayla-glick, line 37: Since this is a really old PR and I just saw it, I went ahead and made the change
https://github.com/wowsims/wotlk/commit/02a97c1f348d2e1ab0b4a30dd65014fe67fdfa93

PR #4151, opened 2024-01-11
fixing missing item error in gear picker on initial load

omccabe:
Applying fix suggested by Lucenia


omccabe:
Closing since @NerdEgghead applied the same change in https://github.com/wowsims/wotlk/commit/fdfff45716c2106a8b941411a2df0d20330aa2bc

PR #4175, opened 2024-01-15
Refactor code that builds Settings page icon pickers

kayla-glick:
This refactor is still a WIP as I've finished Buffs/Debuffs but Consumes are still being worked on. However I wanted to get feedback on the new structure before fully committing.

Key Changes
- I'm trying to break out `ui/core/components/icon_inputs.ts` into separate files because of how large it is currently. I've moved all buff/debuffs/world buff definitions into `ui/core/components/inputs/buffs_debuffs.ts` and consumables to `ui/core/components/inputs/consumables.ts`. This file still seems like a good place to keep the various helper functions in there now.
- There was a lack of flexibility in using different pickers for various buffs/debuffs, so I've created a new type `StatOption` to extend the current functionality. To use this, you provide `config` for a given buff/debuff/consume, `picker` to specify the type of picker you want to use, and `stats` to specify the relevant stats the buff applies to. These are then filtered to only include relevant buffs.


ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -0,0 +1,284 @@
...
+	// makeTristateRaidBuffInput({id: ActionId.fromSpellId(1245), impId: ActionId.fromSpellId(14767), fieldName: 'powerWordFortitude', minLevel: 24, maxLevel: 35}),
+	// makeTristateRaidBuffInput({id: ActionId.fromSpellId(1244), impId: ActionId.fromSpellId(14767), fieldName: 'powerWordFortitude', minLevel: 12, maxLevel: 23}),
+	// makeTristateRaidBuffInput({id: ActionId.fromSpellId(1243), impId: ActionId.fromSpellId(14767), fieldName: 'powerWordFortitude', minLevel: 1, maxLevel: 11}),
+	makeTristateRaidBuffInput({id: ActionId.fromSpellId(11767), impId: ActionId.fromSpellId(18696), fieldName: 'bloodPact'}),
+	makeBooleanRaidBuffInput({id: ActionId.fromItemId(10307), fieldName: 'scrollOfStamina'}),
+], 'Stamina');
+
+export const StrengthRaidBuff = InputHelpers.makeMultiIconInput([
+	makeTristateRaidBuffInput({id: ActionId.fromSpellId(25361), impId: ActionId.fromSpellId(16295), fieldName: 'strengthOfEarthTotem', faction: Faction.Horde}),
+	makeBooleanRaidBuffInput({id: ActionId.fromItemId(10310), fieldName: 'scrollOfStrength'}),
```
NerdEgghead, line 54: Make sure to merge in my changes from yesterday that add Horn of Lordaeron to the Str and Agi raid buff options, and move the scrolls to the Agi and Str consumes options instead. Thanks for putting in the faction filters!

kayla-glick: Aah thanks I merged `sod-fork` in earlier but must've missed that

ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -0,0 +1,570 @@
...
+	// makeTristateRaidBuffInput({id: ActionId.fromSpellId(1244), impId: ActionId.fromSpellId(14767), fieldName: 'powerWordFortitude', minLevel: 12, maxLevel: 23}),
+	// makeTristateRaidBuffInput({id: ActionId.fromSpellId(1243), impId: ActionId.fromSpellId(14767), fieldName: 'powerWordFortitude', minLevel: 1, maxLevel: 11}),
+	makeTristateRaidBuffInput({id: ActionId.fromSpellId(11767), impId: ActionId.fromSpellId(18696), fieldName: 'bloodPact'}),
+	makeBooleanRaidBuffInput({id: ActionId.fromItemId(10307), fieldName: 'scrollOfStamina'}),
+], 'Stamina');
+
+// Separate Strength buffs allow us to use boolean pickers for each
+export const StrengthBuffAlliance = withLabel(
+	makeBooleanRaidBuffInput({id: ActionId.fromSpellId(425600), fieldName: 'hornOfLordaeron', faction: Faction.Alliance}),
+	'Strength',
```
NerdEgghead, line 65: Would name this "Str/Agi" or equivalent since it provides both stats.

ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -0,0 +1,570 @@
...
+);
+
+///////////////////////////////////////////////////////////////////////////
+//                                 DEBUFFS
+///////////////////////////////////////////////////////////////////////////
+
+export const MajorArmorDebuff = InputHelpers.makeMultiIconInput([
+	makeBooleanDebuffInput({id: ActionId.fromSpellId(11597), fieldName: 'sunderArmor'}),
+	makeTristateDebuffInput(ActionId.fromSpellId(11198), ActionId.fromSpellId(14169), 'exposeArmor'),
+], 'Armor Penetration');
```
NerdEgghead, line 229: Would probably keep the old name of "Major ArP" or equivalent here, since CoR and FF below are also "Armor Penetration" debuffs, just minor ones that can be stacked unlike the ones in this category.

ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -0,0 +1,570 @@
...
+export const BoonOfBlackfathom = withLabel(
+	makeBooleanIndividualBuffInput({id: ActionId.fromSpellId(430947), fieldName: 'boonOfBlackfathom'}),
+	'Boon of Blackfathom',
+);
+
+///////////////////////////////////////////////////////////////////////////
+//                                 DEBUFFS
+///////////////////////////////////////////////////////////////////////////
+
+export const MajorArmorDebuff = InputHelpers.makeMultiIconInput([
```
NerdEgghead, line 226: missing Homunculi here

ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -0,0 +1,570 @@
...
+			Stat.StatMP5,
+		]
+	},
+] as PickerStatOptions[];
+
+export const DEBUFFS_CONFIG = [
+	// Standard Debuffs
+	{ 
+		config: MajorArmorDebuff,
+		stats: [Stat.StatArmorPenetration],
```
NerdEgghead, line 498: I had changed this to Attack Power in a previous commit, since Armor Penetration doesn't exist as a stat in vanilla (and is something we should probably remove at some point).

ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -0,0 +1,570 @@
...
+		picker: IconPicker,
+		stats: [Stat.StatMP5, Stat.StatIntellect],
+	},
+	{
+		config: HuntersMark,
+		picker: IconPicker,
+		stats: [Stat.StatRangedAttackPower],
+	},
+
+	// Misc Debuffs
```
NerdEgghead, line 564: Missing the Curse of Vulnerability and Ancient Corrosive Poison debuffs that I had previously added

ui/core/components/inputs/consumables.ts
```diff
@@ -0,0 +1,288 @@
...
+
+// TODO: These should be moved to consumes
+export const GiftOfArthas = withLabel(
+	makeBooleanDebuffInput({id: ActionId.fromSpellId(11374), fieldName: 'giftOfArthas'}),
+	'Gift of Arthas',
+);
+export const CrystalYield = withLabel(
+	makeBooleanDebuffInput({id: ActionId.fromSpellId(15235), fieldName: 'crystalYield'}),
+	'Crystal Yield',
+);
```
NerdEgghead, line 96: I don't think these two should go in consumes, since usually you're modeling someone else in the raid applying these debuffs rather than yourself. For example, DPS sims will often want to check the impact of GoA, but it's the tank who's actually using that consumable.

ui/core/components/individual_sim_ui/consumes_picker.ts
```diff
@@ -193,23 +135,23 @@ export class ConsumesPicker extends Component {
...
-		if (includeAgi){
-			buildIconInput(physicalConsumesElem, this.simUI.player, IconInputs.AgilityBuffInput);
+		if (includeAgi) {
+			buildIconInput(physicalConsumesElem, this.simUI.player, ConsumablesInputs.makeAgilityConsumeInput(ConsumablesInputs.AGILITY_CONSUMES_CONFIG));
 		}
-		if (includeStr){
-			buildIconInput(physicalConsumesElem, this.simUI.player, IconInputs.StrengthBuffInput);
+		if (includeStr) {
+			buildIconInput(physicalConsumesElem, this.simUI.player, ConsumablesInputs.makeStrengthConsumeInput(ConsumablesInputs.STRENGTH_CONSUMES_CONFIG));
 		}
```
NerdEgghead, line 143: Missing Bogling Root here

ui/enhancement_shaman/sim.ts
```diff
@@ -125,7 +125,8 @@ const SPEC_CONFIG = registerSpecConfig(Spec.SpecEnhancementShaman, {
 	rotationInputs: ShamanInputs.EnhancementShamanRotationConfig,
 	// Buff and Debuff inputs to include/exclude, overriding the EP-based defaults.
 	includeBuffDebuffInputs: [
-		IconInputs.MP5Buff,
+		IconInputs.BlessingOfWisdom,
+		IconInputs.ManaSpringTotem,
```
NerdEgghead, line 129: If I'm understanding the changes correctly, these no longer live in IconInputs right?

ui/core/components/inputs/buffs_debuffs.ts
NerdEgghead: Actually now that you've separated it out by faction anyway, it would be even better to put Might and Horn in their own field, since they're exclusive with one another.

kayla-glick: Oh nice I didn't realize that. I can do that then 👍

ui/enhancement_shaman/sim.ts
kayla-glick: Yeah I had been planning to rip all of these out later and solely using EP stats unless there's a reason that wouldn't work? I just haven't gotten that far yet

NerdEgghead: I think the reason `includeBuffDebuffInputs` exists is so that sims have the option of limiting their `epStats` to a smaller subset, while still keeping the full set of buff options. For example, computing EPs for Intellect and MP5 can often be pretty unreliable and overly sensitive to inputs for specs like Hunter or Enhancement Shaman, so the sim authors may want to exclude them to speed up EP calculations and reducing user error. But those specs would still want to allow the corresponding buff/debuff options in their sims though.

kayla-glick: Ah, I see. That's good to know. I can revert the changes then so that those are still used

ui/core/components/individual_sim_ui/consumes_picker.ts
```diff
@@ -221,73 +198,123 @@ export class ConsumesPicker extends Component {
...
+			</div>
+		`;
+
+		this.rootElem.appendChild(fragment.children[0] as HTMLElement);
+		const petConsumesElem = this.rootElem.querySelector('.consumes-pet') as HTMLElement;
+
+		this.simUI.individualConfig.petConsumeInputs.map(iconInput => buildIconInput(petConsumesElem, this.simUI.player, iconInput));
+	}
+
+	private buildPickers({containerElem, changeEmitters, options, parentElem}: {
```
kayla-glick, line 285: This function allows the pickers to be rebuilt dynamically when the relevant emitters fire while also showing / hiding pickers and the outer containers based on what is shown. This could be cleaned up by moving the additional level/faction/equipment filtering to `relevantStatOptions` but the lift is quite a bit bigger since I'd have to redo pretty much all of the buffs/debuffs helper functions.

ui/core/components/individual_sim_ui/consumes_picker.ts
```diff
@@ -41,144 +35,122 @@ export class ConsumesPicker extends Component {
...
+			containerElem: rowElem,
+			options: [
+				{
+					getConfig: () => ConsumablesInputs.makeMainHandImbuesInput(
+						relevantStatOptions(ConsumablesInputs.WEAPON_IMBUES_MH_CONFIG, this.simUI),
+						'Main-Hand',
+					),
+				},
+				{
+					getConfig: () => ConsumablesInputs.makeOffHandImbuesInput(
```
NerdEgghead, line 115: Will this be appropriately hidden if no off-hand weapon is equipped?

ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -0,0 +1,606 @@
...
+	},
+	{
+		config: StaminaBuff,
+		picker: MultiIconPicker,
+		stats: [Stat.StatStamina]
+	},
+	{
+		config: StrengthBuffAlliance,
+		picker: MultiIconPicker,
+		stats: [Stat.StatStrength]
```
NerdEgghead, line 337: would do [Stat.StatStrength, Stat.StatAgility, Stat.StatAttackPower] here

ui/core/components/inputs/buffs_debuffs.ts
```diff
@@ -0,0 +1,606 @@
...
+	// makeTristateRaidBuffInput({actionId: ActionId.fromSpellId(1243), impId: ActionId.fromSpellId(14767), fieldName: 'powerWordFortitude', minLevel: 1, maxLevel: 11}),
+	makeTristateRaidBuffInput({actionId: ActionId.fromSpellId(11767), impId: ActionId.fromSpellId(18696), fieldName: 'bloodPact'}),
+	makeBooleanRaidBuffInput({actionId: ActionId.fromItemId(10307), fieldName: 'scrollOfStamina'}),
+], 'Stamina');
+
+// Separate Strength buffs allow us to use boolean pickers for each
+export const StrengthBuffAlliance = InputHelpers.makeMultiIconInput([
+	makeTristateIndividualBuffInput({actionId: ActionId.fromSpellId(25291), impId: ActionId.fromSpellId(20048), fieldName: 'blessingOfMight', faction: Faction.Alliance}),
+	makeBooleanRaidBuffInput({actionId: ActionId.fromSpellId(425600), fieldName: 'hornOfLordaeron', faction: Faction.Alliance}),
+], 'Strength / Agility');
```
NerdEgghead, line 67: After thinking about this more, I think we should name this `PaladinPhysicalBuff` for maximum clarity, and then make the label `Paladin Physical`, which isn't more characters than `Strength / Agility` and is more clear.

ui/core/components/inputs/consumables.ts
```diff
@@ -0,0 +1,297 @@
...
+	{ config: JujuPower, stats: [Stat.StatStrength] },
+	{ config: ElixirOfGiants, stats: [Stat.StatStrength] },
+	{ config: ElixirOfOgresStrength, stats: [Stat.StatStrength] },
+	{ config: ScrollOfStrength, stats: [Stat.StatStrength] },
+] as ConsumableStatOption<StrengthBuff>[];
+
+export const makeStrengthConsumeInput = makeConsumeInputFactory({consumesFieldName: 'strengthBuff'});
+
+// Other
+export const BoglingRootDebuff = makeBooleanConsumeInput({actionId: ActionId.fromItemId(5206), fieldName: 'boglingRoot'});
```
NerdEgghead, line 203: Extremely minor, but should be `BoglingRootBuff` rather than `BoglingRootDebuff`

ui/core/components/inputs/buffs_debuffs.ts
kayla-glick: Do you know if this applies to Ranged attack power too?

ui/core/components/individual_sim_ui/consumes_picker.ts
kayla-glick: Ah good point!

ui/core/components/inputs/buffs_debuffs.ts
kayla-glick: added `[Stat.StatStrength, Stat.StatAgility, Stat.StatAttackPower, Stat.StatRangedAttackPower]` but let me know if ranged AP is not needed

ui/core/components/individual_sim_ui/consumes_picker.ts
kayla-glick: Implemented logic for both mainhand and offhand imbues and added the `gearchangeemitter`. Confirmed that equipping/unequipping maindhands and offhands showed/hid the pickers

PR #4184, opened 2024-01-16
updating default Retribution APL

Surveillant0:
DS > CS rotation (aka High-roll) as the new default
includes a condition to prio CS higher if Libram duration is low
for people without the Libram, the rotation will end up doing CS > DS instead, unless that action is disabled
however CS > DS is still a perfectly fine and optimal rotation anyway
also includes cancelaura action for Shadowmourne but this is disabled by default, as it is currently bugged in-game

### DIFF
--- a/ui/retribution_paladin/apls/default.apl.json
+++ b/ui/retribution_paladin/apls/default.apl.json
@@ -1,16 +1,21 @@
 {
-    "type": "TypeAPL",
-    "prepullActions": [
-      {"action":{"castSpell":{"spellId":{"otherId":"OtherActionPotion"}}},"doAtValue":{"const":{"val":"-1s"}}}
-    ],
-    "priorityList": [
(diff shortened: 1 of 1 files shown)


Surveillant0:
not sure if I need to fix a bug somewhere or if I need to update expected results, first time updating an APL

kayla-glick:
Hey @Surveillant0! Looks like the update is making the tests fail (which isn't surprising). Since the numbers went up I assume that's totally fine, so you can run `make update-tests` then commit those changes and it should be good!

Surveillant0:
> Hey @Surveillant0! Looks like the update is making the tests fail (which isn't surprising). Since the numbers went up I assume that's totally fine, so you can run `make update-tests` then commit those changes and it should be good!

thanks I will do this as soon as I am able to! appreciate the reply

PR #4193, opened 2024-01-17
SoD: Base stats and crit scaling

FelixPflaum:
Work on base stats and crit.

I got racial offsets ingame using lvl 1 warriors. Base stats are filled from pserver DBs I have on my machine. Testing it against what I see ingame on multiple characters it seems to fit 1:1 now.

Base crit data was collected by using a WA that calculates it and crit per agi/int. I did that with a lvl 1 of each class and also lvl 25 where possible. For the latter I therefore also already have values for the agility and intelligence multiplier.

Considering that the 1.15 client does not ship with crit tables like the 3.4 version, getting those values manually like that seems to be the only way.

Progress:
- [x] Racial offsets
- [x] Base stats
- [x] Base crit
- [x] Level 60 agi/int multipliers (from RatingBuster addon)

Agi/Int multiplier for lvl 25:
- [x] Warrior
- [x] Paladin
- [x] Hunter
- [x] Rogue
- [x] Priest
- [x] Shaman
- [x] Mage
- [x] Warlock
- [x] Druid

Collected can also be seen here: https://docs.google.com/spreadsheets/u/1/d/e/2PACX-1vTGH4N9hc3hHGpsrJ1qHScmIebun7R2Djs9YlTFE6ncErLJY24qX2R3J5aBEkb2drq1UjNt4v_GGRka/pubhtml#


FelixPflaum:
Base values and values for level 25 should be all correct. Hoped that agi/int crit values are just a 2nd degree polynomial, sadly they aren't. Values for level 40 and 50 need to be checked ingame later.

PR #4196, opened 2024-01-20
SOD: explain canRip logic

hotslice:
I want to work more with canRip, but first I will try to explain the logic in the canRip function, and break down the return block into something more manageable. This should make it easier to understand future PRs which will add more logic here.


hotslice:
Closing per discussion with NerdEgghead - I will bundle the functional changes in with the next PR

PR #4205, opened 2024-01-28
WIP: Sim talents in bulk sim

ToxicKevinFerm:
Sorry if I'm making a mess, first time contributing, looking for feedback.

I thought about where to put the talent sims. My use case would be for example simming two set bonuses against eachother, but the set bonus requires a talent swap to maximise the dps.

So for now I've put it in the loop where we add the valid sims. There's still some issues, mainly in the GUI and we sim unnecessary talent loadouts right now. We shouldn't have to sim the currently loaded talent spec even if you pick it, because the bulksim does that automatically.

I think it's also important to note that this obviously multiplies the amount of sims that are needed for a result and so it only does the sims when activated. 

It's not very nicely visualised in the UI either, and you probably want to equip the talent loadout as well on the button press.

### DIFF
--- a/sim/core/bulksim.go
+++ b/sim/core/bulksim.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"reflect"
 	"runtime"
 	"runtime/debug"
 	"sort"
@@ -197,6 +198,10 @@ func (b *bulkSimRunner) Run(pctx context.Context, progress chan *proto.ProgressM
 					for _, talent := range talentsToSim {
 						sr := goproto.Clone(substitutedRequest).(*proto.RaidSimRequest)
 						cl := *changeLog
+						if sr.Raid.Parties[0].Players[0].TalentsString == talent.TalentsString && reflect.DeepEqual(talent.Glyphs, sr.Raid.Parties[0].Players[0].Glyphs) {
+							continue
+						}
+
 						sr.Raid.Parties[0].Players[0].TalentsString = talent.TalentsString
 						sr.Raid.Parties[0].Players[0].Glyphs = talent.Glyphs
 						cl.TalentLoadout = talent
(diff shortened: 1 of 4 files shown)


zku:
Thanks a lot for sending this! To safe some work patching the changes in, could you show a screenshot of how the settings look like, how different talent loadouts are specified, and how the result list from bulk looks like with talent loadouts enabled?

ToxicKevinFerm:
Screenshot_45
Screenshot_49
Screenshot_48
Screenshot_47
Screenshot_46
Here's some screenshots of different views @zku . 
Basically your talent loadouts are the saved talents you define in the talent section. I haven't figured out if I should listen to some events to update the saved talents etc, or what. And the view when youre both simming items and talents at the same time could look better.

ToxicKevinFerm:
Screenshot_50
Updated styling slightly

sim/core/bulksim.go
```diff
@@ -184,7 +186,29 @@ func (b *bulkSimRunner) Run(pctx context.Context, progress chan *proto.ProgressM
...
+			if !b.Request.BulkSettings.SimTalents {
+
+			} else {
+				var talentsToSim = b.Request.BulkSettings.GetTalentsToSim()
+
+				if len(talentsToSim) > 0 {
+					for _, talent := range talentsToSim {
+						sr := goproto.Clone(substitutedRequest).(*proto.RaidSimRequest)
+						cl := *changeLog
+						if sr.Raid.Parties[0].Players[0].TalentsString == talent.TalentsString && reflect.DeepEqual(talent.Glyphs, sr.Raid.Parties[0].Players[0].Glyphs) {
```
zku, line 201: It's been quite a while since I have written / worked on the sim, so please take the comment(s) with a grain of salt. Does this first comparison depend on the order of the specified talent loadouts, or does this always remove duplicates? I.e. if the standard settings use loadout B and the bulk settings specify loadouts [A, B] (in that order), does this comparison actually prevent the duplicate sim?

For the second comparison, can you use proto.Equal? AFAIK reflect.DeepEqual depends on implementation details on protoc for Go, and the generated structs hold things like mutexes and other fields that you probably don't want to compare. https://pkg.go.dev/google.golang.org/protobuf/proto#Equal

zku:
Just one comment above, otherwise looks good to me. Thanks again for working on this!

sim/core/bulksim.go
ToxicKevinFerm: Your first part I'm unsure of, I normally write .NET to be honest, but it does prevent the duplicate sim in the GUI condition. When substitutedRequest is created it uses the base request talents, so whatever is your base simRequest if I understand it correctly.

I changed the DeepEqual to proto.Equal, will verify that it works.

ToxicKevinFerm: `proto.Equal` does work yep, you're right that it seems like a better comparison.

PR #4206, opened 2024-01-28
WIP: Remove legacy rotation code from sod-fork

kayla-glick:


sim/core/cast.go
```diff
@@ -204,10 +199,12 @@ func (spell *Spell) makeCastFunc(config CastConfig) CastSuccessFunc {
 			spell.Unit.SetGCDTimer(sim, sim.CurrentTime+effectiveTime)
 		}
 
+		// TODO: Fix with removal of ChannelTime?
 		if (spell.CurCast.CastTime > 0 || spell.CurCast.ChannelTime > 0) && spell.Unit.Moving {
```
kayla-glick, line 203: Unsure if we should just remove these completely? These are unique to `sod-fork`

sim/druid/feral/rotation.go
```diff
@@ -43,7 +43,6 @@ func (cat *FeralDruid) OnGCDReady(sim *core.Simulation) {
 	// Replace gcd event with our own if we casted a spell
 	if !cat.GCD.IsReady(sim) {
 		nextGcd := cat.NextGCDAt()
-		cat.DoNothing()
```
kayla-glick, line 46: Can this file be removed? It contains powershift logic

ui/core/individual_sim_ui.ts
```diff
@@ -393,20 +391,6 @@ export abstract class IndividualSimUI<SpecType extends Spec> extends SimUI {
 		//this.simHeader.addExportLink("CLI", _parent => new Exporters.IndividualCLIExporter(this.rootElem, this), true);
 	}
 
-	applyDefaultRotation(eventID: EventID) {
```
NerdEgghead, line 396: This was an intentional add on my part to allow individual sims to specify the simple interface as what users see by default if desired. Should still work with the new code as long as the individual sim config supports optional fields within the defaults section for "rotationType?" and "simpleRotation?". I'm happy to re-commit it separately though if that's easier.

NerdEgghead (commented):
Would just double check that Feral sim/UI load properly with these changes since it's the one sim using the Simple interface currently, should be getting around 300 DPS with all defaults. Otherwise looks good!

ui/core/individual_sim_ui.ts
kayla-glick: The feral sim still shows the simple rotation and changing options does affect results, too.

image

kayla-glick: And with full buffs I get ~306 dps

image

NerdEgghead: Yeah that commit wouldn't change results at all, just which interface is shown when loading defaults.

PR #4276, opened 2025-06-10
Fix hunter 4pT10 AP multiplier

dodoels:
## Description

image
According to the databse, hunter's 4pT10 bonus should gain 20% AP instead of 10%. 

This can be verified in WCL log https://classic.warcraftlogs.com/reports/6NWxBQbwMrtvqcX8?fight=11&type=resources&source=38&view=events&spell=1001 at 0:35 the first proc of 4pT10, AP increased from `10288` to `12345`, with delta of `12345-10288 = 2057` which is roughly `2057/10288=19.9999%`

## Result

image

with preset config and p4 preset


dodoels:
should the sim results be updated in the test suites?

PR #4278, opened 2025-06-20
Add Maghia's Misguided Quill to Affliction APL

dodoels:
## Summary

`Maghia's Misguided Quill` is not working and has 0% uptime when simming Affliction because the trinket was missing in the APL. Affliction warlock seems to use the manual trinket control while demo/destro uses `autocastOtherCooldowns` . This is only reproducible with Affliction, other spec and talent work. 

- [ ] make the trinket active in sim (in priority list and bigsnap)
- [ ] reformat the APL

38a8eacd94ad27b17882982f32985d6

## After the fix

Misguided Quill is the better Living Flame so the behavior should be identical

image

DPS gain (p4 default with trinket 2)

image

### DIFF
--- a/ui/warlock/apls/affliction.apl.json
+++ b/ui/warlock/apls/affliction.apl.json
@@ -1,54 +1,336 @@
 {
-    "type": "TypeAPL",
-    "prepullActions": [
-        {"action":{"triggerIcd":{"auraId":{"spellId":71636}}},"doAtValue":{"const":{"val":"-105s"}}},
-        {"action":{"triggerIcd":{"auraId":{"spellId":72416}}},"doAtValue":{"const":{"val":"-65s"}}},
-        {"action":{"triggerIcd":{"auraId":{"spellId":55637}}},"doAtValue":{"const":{"val":"-65s"}}},
-        {"action":{"triggerIcd":{"auraId":{"spellId":75473}}},"doAtValue":{"const":{"val":"-55s"}}},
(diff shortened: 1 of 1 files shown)


ui/warlock/apls/affliction.apl.json
```diff
@@ -1,54 +1,336 @@
...
+			}
+		},
+		{
+			"action": {
+				"condition": { "or": { "vals": [{ "cmp": { "op": "OpGt", "lhs": { "remainingTime": {} }, "rhs": { "const": { "val": "138" } } } }] } },
+				"castSpell": { "spellId": { "itemId": 48722 } }
+			}
+		},
+		{
+			"action": {
```
dodoels, line 72: added

ui/warlock/apls/affliction.apl.json
```diff
@@ -1,54 +1,336 @@
...
+				"sequence": {
+					"name": "big snap",
+					"actions": [
+						{ "castSpell": { "spellId": { "otherId": "OtherActionPotion" } } },
+						{ "castSpell": { "spellId": { "itemId": 45466 } } },
+						{ "castSpell": { "spellId": { "spellId": 33697 } } },
+						{ "castSpell": { "spellId": { "spellId": 54758 } } },
+						{ "castSpell": { "spellId": { "itemId": 45148 } } },
+						{ "castSpell": { "spellId": { "itemId": 48722 } } },
+						{ "castSpell": { "spellId": { "itemId": 50357 } } }
```
dodoels, line 123: added

